PR #2, opened 2021-01-27
Firebase

maryseo070:
- Install firebase, mobx, mobx-react
- Connect fan-fix to firebase
- Create .env.dev and use env vars for firebase configuration
- Add bare user modelling to test out connection


packages/fan-fix/index.html
```diff
@@ -5,6 +5,8 @@
     <title><%= htmlWebpackPlugin.options.title %></title>
   </head>
   <body>
+    <script src="https://www.gstatic.com/firebasejs/8.2.4/firebase-app.js"></script>
+    <script src="https://www.gstatic.com/firebasejs/8.2.4/firebase-firestore.js"></script>
```
mattryavec, line 9: These are redundant, but could be the way to go. Currently, the firebase library is getting built into our webpack bundle. That said, we _might_ switch to using these and exclude firebase from our bundle.

packages/fan-fix/AppInjector.ts
```diff
@@ -0,0 +1,36 @@
...
+      authDomain: process.env.FIREBASE_AUTH_DOMAIN,
+      databaseURL: process.env.FIREBASE_DATABASE_URL,
+      projectId: process.env.FIREBASE_PROJECT_ID,
+      storageBucket: process.env.FIREBASE_STORAGE_BUCKET,
+      messagingSenderId: process.env.FIREBASE_MESSAGING_SENDER_ID,
+      appId: process.env.FIREBASE_APP_ID,
+    });
+  } else {
+    firebase.app();
+ }
```
mattryavec, line 25: Reason for the else clause?

packages/fan-fix/index.html
maryseo070: Ah ok should I remove it from the package.json then and keep these?

packages/fan-fix/AppInjector.ts
maryseo070: oh i was getting a compiler error before without the condition but after it seems to work fine now without it. Maybe it fixed from removing the app container

mattryavec (approved):

PR #31, opened 2021-02-08
Screen 3 (Part 1 of 2 - top section only)

maryseo070:
Creates top section of auth screen with form, button, and captions
- Creates StyledText, StyledTextInput, StyledButton, and View components
- imports fonts


packages/fan-fix/AuthScreen.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <StyledTextInput type="text" placeholder="Name" onChange={(event) =>  this.onInputChange(event, this.name)}/>
+          <StyledTextInput type="text" placeholder="Email" onChange={(event) => this.onInputChange(event, this.email)}/>
+          <StyledTextInput type="password" placeholder="Password" onChange={(event) => this.onInputChange(event, this.password)}/>
+          <StyledText variant="caption">Must be at least 8 characters</StyledText>
+
+          <StyledButton onClick={() => {}} variant="primary" styles={style.verticalMargin}>Sign Up</StyledButton>
+          <StyledText variant="caption" styles={{...style.disclaimer, ...style.verticalMargin}}>
+            You must be at least 13 years old to register. By tapping Sign up, or Sign up with Twitter, Facebook,
+            or Google, you agree to our 
+            <StyledButton variant="link" onClick={() => {}} styles={style.link}>Terms of Use</StyledButton> and
```
mattryavec, line 66: Why variant link and style.link? I'd think the variant would cover the font size!

packages/fan-fix/AuthScreen.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { 
+  StyledButton, 
+  StyledText,
+  StyledTextInput, 
+} from "./ui/controls";
+import { View } from "./ui/controls/View";
+
+const style = StyleSheet.create({
+  root: {
+    marginHorizontal: "24px",
```
mattryavec, line 15: Just `24`? Match what you're doing in verticalMargin below? (I definitely like using raw numbers for pixel sizes)

packages/fan-fix/ui/controls/Stylable.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+type CSSInputType = StyleDeclarationValue | false | null | void;
+
+export type StylesProp = CSSInputType | CSSInputType[];
+
+export interface StylableProps {
+  styles?: StylesProp;
+}
+
+// export function mergeStyles(...stylesProps: StylesProp[]): CSSInputType[] {
```
mattryavec, line 11: Why you no use mergeStyles??

maryseo070: oh i didnt think it was necessary -  i saw on aphrodite docs you can just pass in multiple styles to css() so i just used that
image
image

mattryavec: You certainly can! But the cool thing about mergeStyles is that it makes things work like the `style` prop in ReactNative. That means you can do conditional stuff (`isEnabled && style.enabled`) along with proper handling of multiple styles when an array gets passed in, as the `css` function expects a raw style)

mattryavec: I DON'T WRITE UNNECESSARY CODE MARY! 😂

maryseo070: LOL i was gonna push back and say the css function allows conditions buuuut looking closer i can see how the arrays would be useful

mattryavec: Shiiiittt! Way to keep me honest! Good thing I added that second point!

mattryavec (approved):
KILLER

PR #32, opened 2021-02-09
Screen 3 (Part 2)

maryseo070:
- Add bottom section of screen 3 (third party buttons and link to log in)
- Create Screen component that controls padding based on variant type stack (timeline post stacking) vs page (auth screen)
- Comment out media queries for now


packages/fan-fix/AuthScreen.tsx
```diff
@@ -1,9 +1,11 @@
 import React, { ReactElement } from "react";
-import { StyleSheet, css } from "aphrodite";
+import { StyleSheet } from "aphrodite";
 import { observable } from "mobx";
 import { observer } from "mobx-react";
 import { Inject } from "@not-the-droids/exco-ts-inject";
+import { Palette } from "./ui/style/Palette";
 import { 
+  Screen,
```
mattryavec, line 8: Didn't notice your `index.ts` in `ui/controls` initially. We typically reserve an index.ts for times when we have a truly library/package. What you have thus far is simply a subfolder of the `fan-fix` package, so an index.ts here could lead to some confusion (e.g., notice your View import on line 13 where it's importing directly from the View.ts file instead of your index!)

I do think it makes sense, however, to have a library for these controls for the Admin side of things (why re-invent that?) I think extraction to a top-level package (like the data model lib) would be a prudent step.

packages/fan-fix/ui/controls/View.tsx
```diff
@@ -4,17 +4,11 @@ import { mergeStyles, StylableProps } from "./Stylable";
...
 interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {}
 
-const localStyles = StyleSheet.create({
-  root: {
-    position: "relative",
-    display: "flex",
-    flexDirection: "column",
-  },
-});
+const localStyles = StyleSheet.create({});
```
mattryavec, line 7: If y'ain't usin' it, delete it!

packages/fan-fix/ui/controls/StyledText.tsx
```diff
@@ -11,6 +11,7 @@ type Variant = "body" | "caption";
 
 const variantStyles = StyleSheet.create<Record<Variant, void>>({
```
mattryavec, line 12: Wait, `void`? At the very least this should be `any`, but you should be able to use the `StylesProp` (or `StyleDeclarationValue` from aphrodite)

packages/fan-fix/AuthScreen.tsx
maryseo070: oh this is good to know, i always thought it was a good idea to use an index whenever you're exporting a lot of files from a directory.

packages/fan-fix/AuthScreen.tsx
```diff
@@ -45,26 +58,38 @@ export class AuthScreen extends React.Component {
...
+      <Screen variant="page">
+        <View styles={style.root}>
+          {/* TODO ADD REAL NAVBAR */}
+          <View style={{ height: 85 }}>
+            INSERT NAV BAR 
+          </View>
+          <View styles={[style.formContainer, style.paddingBottom, style.marginBottom]}>
+            <StyledTextInput type="text" placeholder="Name" onChange={(event) =>  this.onInputChange(event, this.name)}/>
+            <StyledTextInput type="text" placeholder="Email" onChange={(event) => this.onInputChange(event, this.email)}/>
+            <StyledTextInput type="password" placeholder="Password" onChange={(event) => this.onInputChange(event, this.password)}/>
```
mattryavec, line 70: Do these onChange events actually work? You're passing in the string _value_ here, not the key. This should be `"password"` and not `this.password`, yes?

mattryavec (approved):

packages/fan-fix/AuthScreen.tsx
maryseo070: ah no lol discovered this on my upcoming pr -  i changed i back to (event) => this.password = event.target.value

maryseo070: actually i might as well update it in this one

packages/fan-fix/AuthScreen.tsx
mattryavec: I mean, it's a matter of taste and preference really. But this has been our convention so far (heh, well, at least since I've been here! 😂 ). But I do find it to be a nice general rule. Otherwise, you get into an arbitrary judgement call for what "a lot of files" is!

PR #33, opened 2021-02-15
Email auth

maryseo070:
- Add email verification page
- Add email link verification function
- Change mobx versions
- Add history injectable


packages/fan-fix/EmailVerificationScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <StyledText>Go find it and tap the link to continue!</StyledText>
+          </View>
+          <StyledButton variant="link" onClick={this.resendEmail}>Resend Email</StyledButton>
+          {
+            this.showEmailInput && (
+              // TODO check with haag on this addition
+              <View styles={style.inputContainer}>
+                <StyledText variant="caption">Please verify your email address</StyledText>
+                <StyledTextInput type="text" placeholder="Email" onChange={(event) => this.email = event.target.value}></StyledTextInput>
+                <StyledButton onClick={this.verifyEmailLink}>Verify Email</StyledButton>
```
mattryavec, line 113: Not verify, but resend, yes?

packages/fan-fix/EmailVerificationScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+          this.showEmailInput = true;
+        }
+      } catch(err) {
+        console.error(err); // todo add error state toast
+      }
+    }
+  }
+
+  resendEmail = async() => {
+    const email = window.localStorage.getItem("emailForSignin");
```
mattryavec, line 90: Whoa... shouldn't this be using `this.email` to _set_ the new value in localStorage?

packages/fan-fix/EmailVerificationScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+  async componentDidMount() {
+    try {
+      await this.verifyEmailLink();
+    } catch(err) {
+      console.error(err);
+    }
+  }
+
+  verifyEmailLink = async() => {
+    let email = window.localStorage.getItem("emailForSignIn") || this.email;
```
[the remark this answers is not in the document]
maryseo070: hm what if they switch devices?

packages/fan-fix/EmailVerificationScreen.tsx
maryseo070: ah this is where i was a little unsure what to do to be honest - i added this case because if they switch devices or open with incognito or something and there is no email in the localStorage, they may have to type it in again and  verify but that flow would be super annoying. Do you have a different suggestion?

packages/fan-fix/EmailVerificationScreen.tsx
maryseo070: lol so in my head the resendEmail would be if they never got the email for some reason and they were waiting on this page.

So the flow would be auth screen (input email) --> sends email link and redirects to Email Verificaiton Screen --> they go to their email and click the link (takes them also to emailVerificationScreen with auth link info in url parms) 

i thought the resend email button would send them another link based on the email from the auth screen
and the verify email button would make them re-enter their email in the case that they switch devices or something and there is no email in local storage. but now that im explaining that to you....i can see that it is confusing and could use some work...

mattryavec:
Re: not having the email address in localstorage.... Let's not concern ourselves with this. If there is no email, redirect them back to the first email entry screen.

And let's not overcomplicate things for ourselves here. SIMPLE FIRST. We've got a long way to go and can't be burning our time on authentication! 😂

packages/fan-fix-ui-controls/style/Palette.ts
```diff
@@ -3,4 +3,5 @@ export const Palette = {
   DarkText: "rgba(0, 0, 0, 0.8)",
   HighlightText: "#FF4C25",
   MutedText: "rgba(0, 0, 0, 0.5)",
+  Black: "202020",
```
mattryavec, line 6: Let's pick a style here :)

packages/fan-fix/EmailVerificationScreen.tsx
```diff
@@ -0,0 +1,94 @@
...
+export class EmailVerificationScreen extends React.Component<Props> {
+  static inject: Inject<ReactElement> = (injector) => {
+    return () => (
+      <EmailVerificationScreen 
+        authenticator={injector.get(Authenticator)()}
+        history={injector.get(HistoryInjectable)()}
+      />
+    );
+  }
+  @observable private email?: string;
```
mattryavec, line 48: Unused now

packages/fan-fix/EmailVerificationScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+      } catch(err) {
+        console.error(err); // todo add error state toast
+      }
+    }
+  }
+
+  resendEmail = async() => {
+    const email = window.localStorage.getItem("emailForSignin");
+    if (email) {
+      await this.props.authenticator.emailLink?.sendVerificationEmail(email, process.env.EMAIL_VERIFICATION_REDIRECT_URL!)
```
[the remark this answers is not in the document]
mattryavec: Still this

mattryavec (approved):

PR #35, opened 2021-02-16
Nav bar

maryseo070:
- Creates navbar component with 5 different variants
- adds icons used in the navbar
PS-- i KNOW there has to be a better way to show which navBar child variant to display than the switch statement (probably using more typescript). but this was my 3rd attempt/change so im VERY open to suggestions.


packages/fan-fix-ui-controls/NavBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { View } from "./View";
+interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
+  title?: string;
+  iconSourceRight?: string;
+  iconSourceLeft?: string;
+  variant?: NavBarVariant;
+  avatarIcon?: string;
+};
+
+export type NavBarVariant = "primary" | "secondary" | "tertiary" | "quaternary" | "quinary";
```
mattryavec, line 15: Holy crap! Seriously, _quinary_ styling????

packages/fan-fix-ui-controls/Screen.tsx
```diff
@@ -1,10 +1,14 @@
...
 import { StylableProps } from "./Stylable";
+import { NavBar, NavBarVariant } from "./NavBar";
 import { View } from "./View";
 
 interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
-  variant?: Variant
+  variant?: Variant,
+  showNavBar?: boolean,
+  navBarTitle?: string,
+  navBarVariant?: NavBarVariant,
```
mattryavec, line 11: How about `navProps?: NavBarProps` If the props exist, show the NavBar

packages/fan-fix/index.css
```diff
@@ -6,6 +6,10 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 }
+.navBar {
+  -webkit-backdrop-filter: blur(10px);
+  backdrop-filter: blur(10px);
+}
```
mattryavec, line 12: Are you using this global class name anywhere? Why not put this in the NavBar component?

packages/fan-fix-ui-controls/NavBar.tsx
maryseo070: haha well its not styling really, but which children to display in the navbar! i looked through the designs and there were 5 different types of navbar (some with the avatar icon, some with the back button, some with ellipsis etc) so i thought it would be best to just define them in the component instead of having each screen define which icon to put on the left and right and what their onClick should be. I think this could definitely be done better. Will think more about it just felt like i was running out of time to spend on one thing so i pushed it

packages/fan-fix/index.css
maryseo070: ah yeah i forgot to add the className to the element, just used aphrodite but i added it here because aphrodite didnt have those styling options so i thought the best place to put them would be here.

packages/fan-fix-ui-controls/NavBar.tsx
```diff
@@ -4,11 +4,11 @@ import { Icon } from "./Icon";
 import { StylableProps } from "./Stylable";
 import { StyledText } from "./StyledText";
 import { View } from "./View";
-interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
+export interface NavBarProps extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
   title?: string;
   iconSourceRight?: string;
   iconSourceLeft?: string;
-  variant?: NavBarVariant;
+  navBarVariant?: NavBarVariant;
```
mattryavec, line 11: Why this rename? `variant` should still work and is nice and consistent :)

packages/fan-fix-ui-controls/NavBar.tsx
mattryavec: Gotcha. Let's definitely revisit when you add some more functionality!

packages/fan-fix-ui-controls/Screen.tsx
```diff
@@ -1,10 +1,13 @@
 import React from "react";
 import { StyleSheet } from "aphrodite";
 import { StylableProps } from "./Stylable";
+import { NavBar, NavBarProps } from "./NavBar";
 import { View } from "./View";
 
 interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
-  variant?: Variant
+  variant?: Variant,
+  navBarTitle?: string,
```
mattryavec, line 9: Just let this get set in `navBarProps`!

mattryavec (approved):
Just a couple minor things. Then looks good!

packages/fan-fix-ui-controls/NavBar.tsx
maryseo070: ahh because screen has a variant. not sure if that would be confusing to pass a variant to screen then within navBarProps also have variant so i renamed to navBarVariant.

mattryavec: I can see the logic! My thought was that since you're now passing these as `navBarProps` you'd still know what it applied to:
```
<Screen
  variant="page"
  navBarProps={{
    variant="quinary"
  }}
/>
```

That said, I also think the NavBar variant will go away when you bring in more functionality ;) So the point is likely moot!

Up to you!!!!

PR #37, opened 2021-02-18
Profile Picture Screens

maryseo070:
- Creates screen 4 and 5 (with conditional rendering based on profile picture URL)
- Add some attributes to App User (this was an accident cuz i initially thought i was starting a different ticket but figured i would just leave it in since we'll need those later)
- Add gradient styling to icon component 
- downgrades history package - version 5 has  issues with re-rendering after history push


packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,5 +1,38 @@
 import { Identifiable } from "@not-the-droids/exco-data-model";
 
+
 export interface AppUser extends Identifiable {
   readonly email?: string;
-}
\ No newline at end of file
+  readonly realName?: RealName;
```
adpoyao, line 6: Is every field here optional? I think username is required, at the very least?

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,5 +1,38 @@
...
+  // readonly followedSubscriptions: ReadonlyArray<Subscription>;
+}
+
+export interface AppUserStatus {
+  suspended?: Date;
+  deactivated?: Date;
+}
+
+export interface CreatorStatus extends AppUserStatus {
+  applied?: Date;
```
adpoyao, line 24: Non-coding related Q: do you know if we need a suspended timestamp as well, or are we tracking it elsewhere outside of CreatorStatus?

packages/fan-fix-ui-controls/IconButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface Props extends React.DetailedHTMLProps<React.ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>,
+  StylableProps {
+    source: string;
+    iconSize?: Size;
+    gradientStyles?: StylableProps,
+    gradientBackgroundType?: GradientType;
+    size?: Size;
+  }
+
+type Size = "small" | "default" | "large";
```
adpoyao, line 15: nit, but alpha sort these?

packages/fan-fix-ui-controls/IconButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+    height: 112,
+    width: 112,
+  },
+  gradient: {
+    marginBottom: 24,
+    backgroundImage: `url(${require("./assets/icons/gradientCircleLarge.png").default})`,
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+    cursor: "pointer",
```
adpoyao, line 45: Should cursor:pointer be applied to all variants of iconButton, or just the gradient variation?

packages/fan-fix-ui-controls/IconButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+    backgroundImage: `url(${require("./assets/icons/gradientCircleMedium.png").default})`,
+  },
+  main: {
+    height: 56,
+    width: 56,
+    backgroundImage: `url(${require("./assets/icons/gradientCircleMain.png").default})`,
+  },
+  large: {
+    height: 112,
+    width: 112,
```
adpoyao, line 37: Does large icontButton variant have a background image?

packages/fan-fix-ui-controls/Screen.tsx
```diff
@@ -12,10 +12,12 @@ interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
 type Variant = "stack" | "page";
 
 const localStyles = StyleSheet.create({
+  root:{ 
+  },
   container: {
+    height: "calc(100vh - 117px)",
```
adpoyao, line 18: I'm wondering where you got the -117px value from?

maryseo070: oh its from the navbar including all the padding and stuff - adds up to 117

packages/fan-fix-ui-controls/IconButton.tsx
maryseo070: oh yeah i put it in gradient - good catch, let me move it

packages/fan-fix-data-model/AppUser.ts
maryseo070: yeah because when I create an appUser/add their profile picture and such, theyre going through the proper sign up flow but they wont have set a username yet.

packages/fan-fix-data-model/AppUser.ts
maryseo070: oh if you'll notice `AppUserStatus` has suspended - and `CreatorStatus` extends it.

packages/fan-fix-ui-controls/Screen.tsx
adpoyao: What do you think about making the navbar height + padding + the sum of those (117) all into a const variable? That way, if one value changes, the rest would update automatically, or at least it appears more clearly where the value comes from since they're dependent of each other.

i.e. 
export const navBarHeight = 100;
export const navBarPadding = 17;
export const totalNavBarHeight = navBarHeight + navBarPadding;

packages/fan-fix-data-model/AppUser.ts
adpoyao: Gotcha!

packages/fan-fix-ui-controls/Screen.tsx
maryseo070: hmm so the navbar is actually 85 px and theres 2 layers of containers/paddings based on how kevin designed it so its children will always align. I like the idea but i feel like it would be more to track since the vertical and horizontal padding are different and there is more than one contianer. for this case my opinion is its fine to leave this as 117 since its unlikely the structure of the navbar will change and if it does we'll have to alter all those areas anyways

packages/fan-fix-ui-controls/Icon.tsx
```diff
@@ -7,7 +7,7 @@ interface Props extends React.HTMLAttributes<HTMLImageElement>, StylableProps {
   size?: Size;
 }
 
-type Size = "default" | "large";
+export type Size = "default" | "large";
```
adpoyao: I would rename this to IconSize so it's clear in other files using this export, and in case other ui control component also requires type Size, we could be more explicit about these.

adpoyao (approved):
Just one additional comment - looks good!

packages/fan-fix/CreateProfileScreen.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  render() {
+    return (
+      <Screen variant="page" navBarProps={{ variant: "quinary", title: "Set profile picture" }}>
+        <View styles={style.root}>
+          <View styles={style.imageContainer}>
+            <IconButton iconSize="small" gradientBackgroundType="large" styles={style.gradient} source="./assets/icons/add.png"/>
+            {
+              this.profilePictureUrl && (
+                <IconButton styles={style.cornerGradient} source={"./assets/icons/trash.png"} iconSize="small" gradientBackgroundType="mainOutline" />
```
mattryavec, line 69: You'll need to require the asset.

Care to make an icon table so we don't sprinkle these requires all around?
```
const icons = {
  trash: require("./assets/...").default,
  ...
}
```

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,5 +1,38 @@
...
+  // readonly createdSubscriptions: ReadonlyArray<Subscription>; // TODO add these after Subscription DM
+  // readonly followedSubscriptions: ReadonlyArray<Subscription>;
+}
+
+export interface AppUserStatus {
+  suspended?: Date;
+  deactivated?: Date;
+}
+
+export interface CreatorStatus extends AppUserStatus {
```
mattryavec, line 23: Eddie and I just had a chat about some of the data model stuff, specifically around inheritance. This looks to be a strong case where we should be favoring _composition over inheritance_ (seriously, Google it, it's worth a read).

Inheritance has it's place, but this will really over-complicate things. Let me know if we need to chat!

packages/fan-fix/CreateProfileScreen.tsx
maryseo070: ah yeah this one is actually located in the fan-fix assets directory so it works wihtout require but I'll move it to ui-controls and make a table

mattryavec (approved):

PR #38, opened 2021-02-18
Screen 7 & 8: "FinishProfileScreen" Design w/out functions

adpoyao:


packages/fan-fix-ui-controls/StyledButton.tsx
```diff
@@ -41,12 +42,15 @@ const variantStyles = StyleSheet.create({
...
+    opacity: 0.25,
   }
 });
 
 export const StyledButton: React.FunctionComponent<Props> = (props) => {
-  const { styles, variant, ...rest } = props;
+  const { disabled, styles, variant, ...rest } = props;
   return (
-    <button {...rest} className={css(mergeStyles(variantStyles[variant || "primary"], styles))}/>
+    <button {...rest} className={css(mergeStyles(disabled && variantStyles.disabled, variantStyles[variant || "primary"], styles))} disabled={disabled} />
```
maryseo070, line 54: the ...rest takes care of the rest of the props including disabled so you dont need to include the disabled={disabled}

maryseo070, line 54: oh also i just noticed mergeStyles is supposed to accept an array of styles so i think it would be good to update - i know it was incorrect before, my bad

mattryavec:
@adpoyao Please write a descriptive title for your PRs instead of just the number of the ticket. I don't want to have to go find the issue to know what you did! And there may be more than one PR to cover a single issue. It's a good habit to get into IMO!

Also I find it constructive to link the issue to your PRs in the side bar, which get you an even better result than using the issue number in your PR title! If you don't want it to auto-close the issue after merge, link it after merge. It makes for a nice record, especially when you're looking at the issue side of things!

packages/fan-fix-ui-controls/StyledButton.tsx
mattryavec: Hold up. It won't! Because you've extract it from props on line 52, `rest` won't contain it! You will have to still explicitly set it as it has been done here.

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -26,24 +49,67 @@ const localStyles = StyleSheet.create({
...
+      <View styles={localStyles.inputContainer}>
+        { type === "textarea" ? (
+            // TODO: Fix Ts error below & make textarea height dynamic - EY
+            <textarea {...textInputProps} className={css(mergeStyles(localStyles.input, localStyles.areaText, style))} value={value}  placeholder={placeholder} />
+          ) : (
+            <input type={type || "text"} {...textInputProps} className={css(mergeStyles(localStyles.input, style))} value={value} placeholder={placeholder}/>
+            )
+        }
+        {/* TODO: Fix Ts error below - EY */}
+        <label className="floatingLabel" style={floatingLabelStyle}>{renderPlaceholder()}</label> {/* TODO: Debug styling here - EY */}
```
maryseo070, line 103: hmm what are these todo's referring to? 
Curious why you're using both className and style - i see you declared `floatingLabel` as a const above but it is not included in the css - was there an issue using the css function?

packages/fan-fix-ui-controls/StyledButton.tsx
maryseo070: oh my bad! didnt think about it being used for styling too

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -26,24 +49,67 @@ const localStyles = StyleSheet.create({
...
+      </View>
+      {
+        (message?.text) && (
+          <View style={{ marginTop: 8 }}>
```
maryseo070, line 107: i think it would be better to use the styles prop instead of inline styling

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  root: {
+    width: "100%",
+    height: "100%",
+  },
+  formContainer: {
+    height: "100%",
+    display: "flex",
+    flexDirection: "column",
+    justifyContent: "space-between",
+
```
maryseo070, line 25: remove extra line

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  // Helper functions:
+  // Username search function is debounced by n seconds to minimize overloading server with api calls.
+  checkExistingUsername = debounce(async (): Promise<void> => {
+    // query this.username
+    await new Promise(resolve => setTimeout(resolve , USERNAME_DEBOUNCE_DURATION)); // TODO: Replace with real API call
+    const usernameExists = true;
+    const type = usernameExists ? "error" : "success";
+    const text = usernameExists ? "This username is taken" : "This username is available";
```
mattryavec, line 87: How about:
```
const message: Message = usernameExists ?
  { type: "error", message: "This username is taken" } :
  { type: "success", message: "This username is available" };
```

Even better, you can make the success and error messages constants :)

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+  onUsernameMessageChange = (message: Message) => {
+    this.usernameMessage = message;
+  }
+
+  // Helper functions:
+  // Username search function is debounced by n seconds to minimize overloading server with api calls.
+  checkExistingUsername = debounce(async (): Promise<void> => {
+    // query this.username
+    await new Promise(resolve => setTimeout(resolve , USERNAME_DEBOUNCE_DURATION)); // TODO: Replace with real API call
```
mattryavec, line 84: Don't want to stub out a fake implementation in the data model?

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+    this.checkExistingUsername();
+  }
+  onBioChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const input = this.handleMaxCharCount(event.target.value, BIO_MAX_CHAR_COUNT);
+    this.bio = input;
+  }
+  onWebsiteLinkChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.websiteLink = event.target.value;
+  }
```
maryseo070, line 75: nit: add spaces between functions plzz

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -2,19 +2,42 @@ import React from "react";
...
 import { mergeStyles, StylableProps } from "./Stylable";
 import { Palette } from "./style/Palette";
+import { StyledText } from "./StyledText"
+import { View } from "./View";
 
+export type MessageType = "success" | "error";
+export interface Message {
+  type?: MessageType,
+  text?: string,
+};
```
mattryavec, line 12: `Message` is going to get pretty overloaded here. Nothing indicates the context for this message. (Our typical notion of a "message" has been a notification of some sort.)

For these types of things, I like to prefix with the name of the main type of the file (in this case, `StyledTextInput`). So I'd go with `StyledTextInputMessage` to keep things nice and unambiguous.

packages/fan-fix-ui-controls/StyledTextInput.tsx
mattryavec, line 107: Extract to localStyles

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -26,24 +49,67 @@ const localStyles = StyleSheet.create({
...
+    return (
+      <>
+        <View>{placeholder}</View>
+        <View>{currentLength}/{maxLength}</View>
+      </>
+    )
+  }
+
+  const messageColor = message?.type === "success" ? localStyles.success : message?.type === "error" ? localStyles.error : null;
+  const floatingLabelStyle = Boolean(value) ? floatingLabel : { display: "none" };
```
mattryavec, line 90: No need to cast to a Boolean. The truthy/falsy will do the exact same thing. Just bare `value` is fine.

mattryavec, line 90: Extract your used in the false case to localStyles

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -26,24 +49,67 @@ const localStyles = StyleSheet.create({
...
+    const currentLength = value ? value.length : 0;
+    return (
+      <>
+        <View>{placeholder}</View>
+        <View>{currentLength}/{maxLength}</View>
+      </>
+    )
+  }
+
+  const messageColor = message?.type === "success" ? localStyles.success : message?.type === "error" ? localStyles.error : null;
```
mattryavec, line 89: I'd do this inline down on line 108. You've already tested for the existence of a message, so this makes a bit more sense (and you needed have the `?` qualifiers). Also, I love having relevant pieces of code close together spatially within a file when possible. I find it improves readability.

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -26,24 +49,67 @@ const localStyles = StyleSheet.create({
...
+  error: {
+    color: Palette.ErrorText,
+  }
+});
 
 export const StyledTextInput: React.FunctionComponent<Props> = (props) => {
-  const { style, placeholder, type, ...rest } = props;
+  const { maxLength, message, placeholder, style, type, value, ...textInputProps } = props;
+
+  const renderPlaceholder = () => {
```
mattryavec, line 76: Why make this a function?

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -26,24 +49,67 @@ const localStyles = StyleSheet.create({
...
+
   return (
-    <div className={css(localStyles.inputContainer)}>
-      <input type={type || "text"} {...rest} className={css(mergeStyles(localStyles.input, style))} placeholder={placeholder}/>
-      <label className="floatingLabel" style={{ display: "none" }}>{placeholder}</label>
-    </div>
+    <>
+      <View styles={localStyles.inputContainer}>
+        { type === "textarea" ? (
+            // TODO: Fix Ts error below & make textarea height dynamic - EY
```
mattryavec, line 96: What are these TS errors? I'd rather not merge code with compile errors

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  onUsernameMessageChange = (message: Message) => {
+    this.usernameMessage = message;
+  }
+
+  // Helper functions:
```
mattryavec, line 80: I find calling things "helper" or "util" functions to be redundant and rather meaningless! All functions are helpful and have utility, wouldn't you say? :)

If you'd like to keep your documentation, perhaps be more specific (e.g., `Username validation`)

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const usernameIsValid = this.usernameMessage.type === "success";
+    return this.username && usernameIsValid && !!this.bio && !!this.websiteLink;
+  }
+
+  render() {
+    const buttonIsDisabled = !this.validateProfileFormInput();
+    return (
+      <Screen variant="page">
+        <View styles={style.root}>
+          {/* TODO ADD REAL NAVBAR */}
```
mattryavec, line 109: I think this is merged now. @maryseo070  ?

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+    paddingBottom: 24,
+  },
+  textInputWrapper: {
+    marginBottom: 8,
+  }
+});
+
+const USERNAME_MAX_CHAR_COUNT = 20;
+const BIO_MAX_CHAR_COUNT = 140;
+const USERNAME_DEBOUNCE_DURATION = 200;
```
mattryavec, line 43: Now I'm getting exceedingly nitpicky, but I find the UPPER_SNAKE_CASING of constants to smack of [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation), a thoroughly unpleasant convention!

What makes these constants so special than the name of any other variable or constant? Just because there in a more outer scope? Are they that much more important? Personally, I like sticking with the standard lower-camel-cased convention you'd use for any other variable.

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+    await new Promise(resolve => setTimeout(resolve , USERNAME_DEBOUNCE_DURATION)); // TODO: Replace with real API call
+    const usernameExists = true;
+    const type = usernameExists ? "error" : "success";
+    const text = usernameExists ? "This username is taken" : "This username is available";
+    this.showUsernameMessage({ type, text });
+  }, 200);
+  showUsernameMessage = (message: Message) => {
+    this.usernameMessage = message;
+  }
+  handleMaxCharCount = (input: string, maxChar: number): string => {
```
mattryavec, line 93: Does the `maxLength` property on the text input not handle this for you? If not, it would be useful to have `StyledTextInput` itself guard against this so you don't need to do this for every bounded length input field.

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  @observable private bio: string = "";
+  @observable private websiteLink: string = "";
+  // TextInput controlled messages
+  @observable private usernameMessage: Message = {};
+
+  onUsernameChange = async (event: React.ChangeEvent<HTMLInputElement>) => {
+    const input = this.handleMaxCharCount(event.target.value, USERNAME_MAX_CHAR_COUNT);
+    this.username = input;
+    if (!this.username) {
+      this.showUsernameMessage({});
```
mattryavec, line 64: Perhaps just allow the message itself to be undefined? Then you could make the fields on `Message` required

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,132 @@
...
+  showUsernameMessage = (message: Message) => {
+    this.usernameMessage = message;
+  }
+  handleMaxCharCount = (input: string, maxChar: number): string => {
+    if (input.length > maxChar) {
+      input = input.slice(0, maxChar);
+    }
+    return input;
+  }
+  validateProfileFormInput = () => {
```
mattryavec, line 99: Make this a `@computed` property!
```
@computed private get isFormValid(): boolean {
  const usernameIsValid = this.usernameMessage.type === "success";
  return this.username && usernameIsValid && !!this.bio && !!this.websiteLink;
}
```

packages/fan-fix-ui-controls/package.json
```diff
@@ -8,6 +8,7 @@
   },
   "dependencies": {
     "@types/react": "^17.0.1",
+    "lodash": "^4.17.20",
```
mattryavec, line 11: Use the specific function libraries to cut down on bundle size: `lodash.debounce`

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -2,19 +2,42 @@ import React from "react";
 import { StyleSheet, css } from "aphrodite";
 import { mergeStyles, StylableProps } from "./Stylable";
 import { Palette } from "./style/Palette";
+import { StyledText } from "./StyledText"
+import { View } from "./View";
 
+export type MessageType = "success" | "error";
+export interface Message {
+  type?: MessageType,
+  text?: string,
```
mattryavec, line 11: Let's make these required instead of optional fields. (See my comment in `FinishProfileScreen` below regarding this!)

packages/fan-fix-ui-controls/package.json
maryseo070, line 11: hiiii eddie haha
so i think maybe @mattryavec can help us decide here but i read this medium article (https://codeburst.io/why-you-shouldnt-use-lodash-anymore-and-use-pure-javascript-instead-c397df51a66) on how lodash can significantly hurt the speed of your app and in many examples its better to use plain JS. My preference would be to use JS but would be good to have a discussion between us all

packages/fan-fix/FinishProfileScreen.tsx
maryseo070: yup! its `navBarProps` under `Screen`

packages/fan-fix-ui-controls/package.json
mattryavec: This article is referring to things that lodash provides (like finding something in an array) that are now provided in the core JS spec (like Array#find). The lodash impls for things like `get` are also notoriously slow if overused.

However, things like `debounce` and `throttle` are still _very_ useful and well-written and there is no built-in equivalent.

maryseo070: ok cool! thanks for clarifing matt! i looked at the source code for debounce and it didnt seem too bad but wasnt sure. and i know we use get a lot in hst and odt so its good to know which ones are quesitonable

packages/fan-fix-ui-controls/StyledTextInput.tsx
adpoyao: I have the todo resolved in a coming PR;
the className `floatingLabel` is used to target the ":focus ~ .floatingLabel"
the `floatingLabel` (which I later renamed to floatingLableStyle) is dynamic and it manages floating labels between empty or valid values. Let me know if you think there's a better way to consolidate them!

packages/fan-fix/FinishProfileScreen.tsx
adpoyao: Good idea!

packages/fan-fix/FinishProfileScreen.tsx
adpoyao: Completely forgotten about it!

packages/fan-fix-ui-controls/StyledTextInput.tsx
adpoyao: Done!

packages/fan-fix-ui-controls/package.json
adpoyao: Thanks! I'll refrain from lodash. In this case, I'm using lodash's `debounce` and I've imported just that function per Matt's comment above - changes to come in my next commit push.

packages/fan-fix-ui-controls/StyledTextInput.tsx
adpoyao: done!

packages/fan-fix-ui-controls/StyledTextInput.tsx
adpoyao: All resolved!

packages/fan-fix/FinishProfileScreen.tsx
adpoyao: Got it! I'd say so, now that you'd pointed it out ;) I'll remove the comment, I think it's rather self explanatory.

packages/fan-fix/FinishProfileScreen.tsx
adpoyao: Ohhh you brought out wikipedia! That's good to know, I actually don't even know where I got this convention from. Point taken! Will amend.

packages/fan-fix-ui-controls/StyledTextInput.tsx
adpoyao: I ended up moving this function outside the component class and exported it to another styled component I created: `StyledTextArea`. Those do similar thing in the UI, but I needed the textarea input box to have dynamic height which required additional logic. I wanted to keep those two separate as to not to muddy the logic. Let me know if that works.

packages/fan-fix-ui-controls/StyledTextInput.tsx
adpoyao: I see what you mean about having tested the existence of a message, I didn't think about that before. Thank you!

packages/fan-fix/FinishProfileScreen.tsx
adpoyao: This is carried over to the `StyledTextInput` 💯

packages/fan-fix-ui-controls/package.json
mattryavec: Still this!

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -0,0 +1,131 @@
...
+      <FinishProfileScreen />
+    );
+  }
+
+  // TextInput controlled values
+  @observable private username: string = "";
+  @observable private bio: string = "";
+  @observable private websiteLink: string = "";
+  // TextInput controlled messages
+  @observable private usernameMessage: StyledTextInputMessage | undefined = undefined;
```
mattryavec, line 65: Nit: For props, I typically prefer the optional `?` syntax (`private usernameMessage?: StyledTextInputMessage`). It's functionally equivalent!

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -1,49 +1,143 @@
...
+    const currentLength = value?.length || 0;
+    return (
+      <>
+        <View>{placeholder}</View>
+        <View>{currentLength}/{maxLength}</View>
+      </>
+    )
+  }
+  
+  const handleMaxCharCount = (input: string, maxChar: number): string => {
```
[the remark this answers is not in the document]
adpoyao: Of course I like to complicate things... forgot about the maxLength attribute.

packages/fan-fix-ui-controls/StyledTextInput.tsx
```diff
@@ -1,49 +1,124 @@
...
+      return <View>{placeholder}</View>;
+    }
+    const currentLength = value?.length || 0;
+    return (
+      <>
+        <View>{placeholder}</View>
+        <View>{currentLength}/{maxLength}</View>
+      </>
+    )
+  }
```
mattryavec, line 102: Similar question as before, why the function? Why not something like:
```
let placeholderElement = <View>{placeholder}</View>
if (maxLength) {
  placeholderElement = (
    <>
      {placeholderElement}
      <View>{currentLength}/{maxLength}</View>
  );
}
```

packages/fan-fix-ui-controls/package.json
```diff
@@ -7,12 +7,20 @@
     "test": "jest"
   },
   "dependencies": {
-    "@types/react": "^17.0.1",
     "mobx": "4.15.3",
     "mobx-react": "5.4.3",
-    "react": "17.0.1"
+    "react-textarea-autosize": "^8.3.1"
```
mattryavec, line 12: Fix the version

mattryavec (approved):
Digging it. Just a couple things, then looks good!

PR #39, opened 2021-02-20
Creation of Subscription data model

adpoyao:


adpoyao:
@maryseo070 I will create a separate PR for typing out Creators after your updated AppUser model merges to master 👍

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,13 @@
...
+import { AppUser } from "./AppUser";
+
+export interface Subscription extends Identifiable {
+  // category: List<Category> // TODO: pending Category - EY
+  creatorId: IdentifiableReference<AppUser>;
+  creatorSuspended: boolean;
+  description: string;
+  price: number;
+  // posts: List<Post> // TODO: pending Post - EY
+  subscribers: List<IdentifiableReference<AppUser>>;
```
adpoyao, line 11: Please check my typing here for subscribers. I'm not sure if a list of id'able reference is the way to go about it. Thanks.

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,13 @@
+import { Identifiable, IdentifiableReference, List } from "@not-the-droids/exco-data-model";
+import { AppUser } from "./AppUser";
+
+export interface Subscription extends Identifiable {
+  // category: List<Category> // TODO: pending Category - EY
+  creatorId: IdentifiableReference<AppUser>;
```
maryseo070, line 6: Is this meant to be a reference or a string that will be the appUser's id?

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,13 @@
...
+  creatorSuspended: boolean;
```
maryseo070, line 7: thinking more about the data model, i dont think this flag is necessary. We should be able to add that __suspended flag to posts in the process of suspending a creator so in my opinion adding this to a subscription would be redundant

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,13 @@
...
+  description: string;
+  price: number;
+  // posts: List<Post> // TODO: pending Post - EY
+  subscribers: List<IdentifiableReference<AppUser>>;
+  subscribersCount: number;
```
maryseo070, line 12: hmmm i think this will need to contain more infeormation than just a number. Not sure if you remember cuz it was a while ago but we're keeping track of the number of subscribers for the creator's analytics section which will break down subscribers by month and then include a total count. so i think it might be better to track counts by month in addition to a total count. what do you think?

packages/fan-fix-data-model/Subscription.ts
maryseo070: oh also dont forget we wont be using observable references or lists because of how large the data could get

packages/fan-fix-data-model/Subscription.ts
maryseo070: Hmm so i think since we'll be saving the subscribers in a subcollection it wont be a list of identifiable references, just a list of AppUser s . But I'll let matt chime in just in case hahah

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -0,0 +1,48 @@
+import { IdentifiableReference } from "@not-the-droids/exco-data-model";
+import {
+  createOptionalDocumentDeserializer,
+  FirebaseObservableReference,
```
maryseo070, line 4: ahh we are not using observable lists/references for FF because the load we're expecting will surpass the realtime limit for observing. So instead of getting refs we can just get the real thing!

packages/fan-fix/FakeAppInjector.ts
```diff
@@ -1,12 +1,15 @@
...
   AppUserModel,
+  SubscriptionModel,
 } from "fan-fix-data-model";
 
 export function createAppInjector(): Injector {
   const injector = new Injector();
   // Add bindings here
-  injector.bind(AppUserModel, FakeAppUserModel)
+  injector.bind(AppUserModel, FakeAppUserModel);
+  injector.bind(SubscriptionModel, FakeSubscriptionModel);
```
maryseo070, line 13: did you want to add the firebase one to the real app injector too?

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Inject, singletonInject } from "@not-the-droids/exco-ts-inject";
+import { Subscription } from "../Subscription";
+import { SubscriptionModel } from "../SubscriptionModel";
+
+const deserializeSubscription = createOptionalDocumentDeserializer<Subscription>((id, data) => {
+  return {
+    id,
+    // category: data.category, // TODO: pending Category - EY
+    creatorId: data.creatorId,
+    creatorSuspended: data.creatorSuspended,
```
maryseo070, line 17: can remove this

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -0,0 +1,48 @@
...
+} from "@not-the-droids/exco-data-model-firebase";
+import { Inject, singletonInject } from "@not-the-droids/exco-ts-inject";
+import { Subscription } from "../Subscription";
+import { SubscriptionModel } from "../SubscriptionModel";
+
+const deserializeSubscription = createOptionalDocumentDeserializer<Subscription>((id, data) => {
+  return {
+    id,
+    // category: data.category, // TODO: pending Category - EY
+    creatorId: data.creatorId,
```
maryseo070, line 16: when we grab subscriptions we'll probably want to show the creator of the subscription (i didnt look too closely at each design page for subscription but i think we will showw at least their name) so it would be good to deserialize the creator data more to how we're expecting to use it on the FE.

packages/fan-fix-data-model/Subscription.ts
adpoyao: Ah yes, the data model calls for the Id so this should've been the string of ID.

packages/fan-fix-data-model/Subscription.ts
adpoyao: I think you're right, I'll remove this creatorSuspended. If there's a use case for it, we could always add it back but yes, looking over the features, suspending posts and creator alone should be sufficient.

packages/fan-fix-data-model/Subscription.ts
adpoyao: I chatted with Matt and this will be updated to `List<AppUser>>`

packages/fan-fix/FakeAppInjector.ts
adpoyao: Ah forgot. Thanks!

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,5 +1,17 @@
 import { Identifiable } from "@not-the-droids/exco-data-model";
+import { Subscription } from "./Subscription";
 
 export interface AppUser extends Identifiable {
   readonly email?: string;
+}
+
+export interface Creator extends AppUser {
+  readonly creatorStatus?: CreatorStatus;
+  readonly createdSubscriptions: ReadonlyArray<Subscription>
```
mattryavec, line 10: Why "created" subscriptions and not just plain ol' subscriptions?

Also this shouldn't this be a `List<Subscription>`? And if so, to be in line with the Reference nomenclature, I like to suffix with -List (e.g., `createdSubscriptionList`) as IMO it more clearly disambiguates it from a regular collection (Array, Map, Set, etc) for which I like to reserve for the plural.

If you all have a different convention you'd like to adopt, however, I'm all ears!

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,12 @@
+import { Identifiable, List, Reference } from "@not-the-droids/exco-data-model";
+import { AppUser } from "./AppUser";
+
+export interface Subscription extends Identifiable {
+  // category: List<Category> // TODO: pending Category - EY
+  creatorRef: Reference<AppUser>;
+  description: string;
+  price: number;
+  // posts: List<Post> // TODO: pending Post - EY
+  subscribers: List<AppUser>;
```
mattryavec, line 10: `subscriberList` if you like my previous suggestion

packages/fan-fix-data-model/fake/FakeSubscriptionModel.ts
```diff
@@ -0,0 +1,20 @@
...
+    return () => new FakeSubscriptionModel();
+  });
+
+  constructor() {
+    super();
+  }
+
+  public getSubscriptionRef(id: string): IdentifiableReference<Subscription> {
+    const ref = new FakeObservableIdentifiableReference<Subscription>(id);
+    return ref;
```
mattryavec, line 18: There is a FakeDataManager you can make use of. @maryseo070 can show you the way....

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -0,0 +1,47 @@
...
+      injector.get(FirestoreInjectable)(),
+    );
+  });
+  private readonly subscriptionCollection: UnifiedFirestoreTypes.CollectionReference;
+
+  constructor(
+    private readonly db: UnifiedFirestoreTypes.Module,
+  ) {
+    super();
+    this.subscriptionCollection = db.collection("Subscription");
```
mattryavec, line 37: @maryseo070 Tell Eddie about the FirebaseDataManager as well!

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
mattryavec: I was talking to Eddie about this this morning. I think it would be useful to have a "fetch" style observable reference and list. It could simply fetch once (i.e., initialize itself) when it becomes observed and reset itself back to uninitialized when it becomes unobserved. The explicit `initialize` would of course also do the fetch. I'd even be open to making this configurable on the existing FirebaseObservableReference/List classes. (e.g., "fetch" mode or "realtime" mode)

packages/fan-fix-data-model/AppUser.ts
adpoyao: I like the convention of adding `List` for things that are lists.
Reason for "created": going by the data model, an app user has a "createdSubscriptionList" and "followedSubscriptionList". Let us know if you think this is appropriate!

packages/fan-fix-data-model/firebase/FirebaseStatisticModel.ts
```diff
@@ -1,45 +1,46 @@
...
+  };
+
+  // STATISTICS
   public getStatisticRef(id: string): IdentifiableReference<Statistic> {
-    const statisticRef = new FirebaseObservableReference( // TODO: Update to FetchableObservableReference - EY
-      this.statisticCollection.doc(id),
-      deserializeStatistic,
-    )
-    return statisticRef;
+    return this.statisticDataManager.getRef(id); // TODO: Update with fetch mode option - EY
```
adpoyao, line 44: I will work on implementing fetchObservableReference in my next PR.

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,5 +1,14 @@
-import { Identifiable } from "@not-the-droids/exco-data-model";
+import { Identifiable, List } from "@not-the-droids/exco-data-model";
+import { Subscription } from "./Subscription";
 
 export interface AppUser extends Identifiable {
   readonly email?: string;
+  readonly creatorStatus?: CreatorStatus;
+  readonly createdSubscriptionList: List<Subscription>
```
maryseo070, line 7: nit: semicolon

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,5 +1,14 @@
...
 export interface AppUser extends Identifiable {
   readonly email?: string;
+  readonly creatorStatus?: CreatorStatus;
+  readonly createdSubscriptionList: List<Subscription>  
+}
+
+export interface CreatorStatus { // TODO: implement "extends AppUserStatus" once merges from Mary's PR - EY
+  applied?: Date;
+  accepted?: Date;
+  rejected?: Date;
```
maryseo070, line 13: in my opinion the `at` more indicates a date i.e.,) appliedAt, acceptedAt, rejectedAt -- i think it would be good to follow that for FF, could you update thesE?

packages/fan-fix-data-model/Statistic.ts
```diff
@@ -0,0 +1,11 @@
+import { Identifiable, Reference } from "@not-the-droids/exco-data-model";
```
maryseo070, line 1: oh do you still need the Category.ts file?

packages/fan-fix-data-model/Statistic.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Day } from "@not-the-droids/exco-data-model-time";
+import { Subscription } from "./Subscription";
+
+export interface Statistic extends Identifiable {
+  subscriptionRef: Reference<Subscription>;
```
maryseo070, line 6: this is a nit but can you use Identifiable Reference? I think there could be cases where we just check the id of the subscription

packages/fan-fix-data-model/Statistic.ts
```diff
@@ -0,0 +1,11 @@
...
+  day: Day,
+  followerCount: number;
+  likeCount: number;
+  viewCount: number;
```
maryseo070, line 10: I'll let @mattryavec comment more on this since im not 100% confident but i was thinking we could have a cumulative totalCount that takes from yesterday's totalCount and adds to it. That way when we want to fetch the total count we dont have try to combine all the totalCounts from the documents and such.

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,13 @@
+import { Identifiable, List, Reference } from "@not-the-droids/exco-data-model";
+import { AppUser } from "./AppUser";
+import { Statistic } from "./Statistic";
+
+export interface Subscription extends Identifiable {
+  // category: List<Category> // TODO: pending Category - EY
+  creatorRef: Reference<AppUser>;
```
maryseo070, line 7: i think its a good policy to use identifiable references because in the case when we just want to see if they have a creatorRef or what that creator ref's id is or something we dont need to call initialize or fetch the whole data, we can just get the id

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -0,0 +1,13 @@
...
+  description: string;
+  price: number;
+  // posts: List<Post> // TODO: pending Post - EY
+  subscriberList: List<AppUser>;
+  statisticList: List<Statistic>;
```
maryseo070, line 12: i dont believe all these fields should be required haha

packages/fan-fix-data-model/SubscriptionModel.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Inject } from "@not-the-droids/exco-ts-inject";
+import { Subscription } from "./Subscription";
+
+export abstract class SubscriptionModel {
+  static inject?: Inject<SubscriptionModel>;
+
+  public abstract getSubscriptionRef(id: string): IdentifiableReference<Subscription>;
+}
+
+export type CreateSubscriptionParams = Subscription & {
```
maryseo070, line 11: hmm i saw your todo about using update params later when needed but actually i think you can just consolidate them! i dont think we need a separate type for create and update, so might as well just use update!

packages/fan-fix-data-model/fake/FakeSubscriptionModel.ts
```diff
@@ -0,0 +1,40 @@
...
+    ...data,
+    id,
+  }))
+
+  constructor(
+    private readonly appUserModel: FakeAppUserModel,
+  ) {
+    super();
+    // this.subscriptionDataManager.insert(this.serializeSubscription({})); // TODO: Insert fake date once it is needed - EY
+  }
```
maryseo070, line 25: lol any reason why you're making a todo for these instead of just stubbing them out? You might as well just create a fake since you're setting everything else up, right?! haha up to you but would be nice so we dont rack up the todos too bad

packages/fan-fix-data-model/fake/FakeSubscriptionModel.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  private serializeSubscription(subscription: CreateSubscriptionParams) { // TODO: Update createSubscriptionParams & created UpdateSubscriptionParams once it is needed. - EY
+    return {
+      ...subscription,
+      creatorRef: this.appUserModel.getAppUserRef(subscription.creatorId),
+    }
+  };
+
+
+  // SUBSCRIPTIONS
```
maryseo070, line 35: i dont think this comment is very necessary hahah

packages/fan-fix-data-model/fake/FakeStatisticModel.ts
```diff
@@ -0,0 +1,33 @@
...
+    // this.statisticDataManager.insert(this.serializedStatistic({})); // TODO: Insert fake date once it is needed - EY
+  }
+
+  private serializeStatistic(statistic: Statistic) { // TODO: Update to new CreateStatisticParams | UpdateStatisticParams once it is needed. - EY
+    return {
+      ...statistic,
+    }
+  };
+
+  // STATISTIC
```
maryseo070, line 28: dont think we need this comment

packages/fan-fix/FakeAppInjector.ts
maryseo070, line 13: i think you forgot to add FakeStatisticModel

packages/fan-fix-data-model/Subscription.ts
adpoyao: Which ones do you think I should remove? I was going off of the Data Model chart

packages/fan-fix-data-model/SubscriptionModel.ts
adpoyao: I think we need to have them separate because Update Param requires the document Id, whereas Create does not!

packages/fan-fix-data-model/fake/FakeSubscriptionModel.ts
adpoyao: Ah okay. I was following the convention of another file, I'll remove this now

packages/fan-fix-data-model/Statistic.ts
adpoyao: hm? I'm not seeing the Category.ts file here.

packages/fan-fix-data-model/fake/FakeSubscriptionModel.ts
adpoyao: My thought was to wait for the data model to first be completed (i.e. Category and Posts) and approved, then whoever once we're working on the functionality, the fake could be implemented to what's been finalized. I see what you're saying too and will write out a fake here now.

packages/fan-fix-data-model/Statistic.ts
mattryavec: I want to keep the daily counts no matter what.

I am open to adding a field to track the cumulative 30-day count, but I don't like that it ties us to a particular window this early in the game.

Fetching 30 documents is a light lift starting out. I say keep it simple for now and we evolve it if it becomes a performance issue.

maryseo070: ahh i meant total as in complete total, not just 30 days - so every day we continue to add the daily total to the cumulative total

mattryavec: For overall totals, I'd suggest maintaining that in a separate document. Perhaps `Statistic/totals`.

Do we need those total numbers?

mattryavec (approved):
Good with it...

PR #40, opened 2021-02-28
Explore Page - First Iteration

adpoyao:
Implements:
- Home Screen + Tab Bar  + Bubble Chart (D3)

Next Iteration to include:
- bubble chart parent height & width: change to dynamic setting
- bubble chart background image
- make svg screen responsive as screen resizes without reload (will check with Kevin if this is needed)
- Update D3 function parameters with typescript types


packages/fan-fix-ui-controls/TabBar.tsx
```diff
@@ -0,0 +1,138 @@
+import React from "react";
+import { StyleSheet } from "aphrodite";
+// import { Link } from "react-router-dom";
```
maryseo070, line 3: still need tthis?

packages/fan-fix-ui-controls/TabBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+const localStyles = StyleSheet.create({
+  navBarContainer: {
+    height: 52,
+    display: "block",
+    left: 0,
+    right: 0,
+    bottom: 0,
+    top: "auto",
+    position: "sticky",
+    paddingBottom: "20px",
```
maryseo070, line 28: can just use 20 instead of the string

packages/fan-fix-ui-controls/TabBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+  activity: require("./assets/icons/icon-activity@2x.png").default,
+  activityFilled: require("./assets/icons/icon-activity-filled@2x.png").default,
+  create: require("./assets/icons/icon-create@2x.png").default,
+  explore: require("./assets/icons/icon-explore@2x.png").default,
+  exploreFilled: require("./assets/icons/icon-explore-filled@2x.png").default,
+  following: require("./assets/icons/icon-following@2x.png").default,
+  followingFilled: require("./assets/icons/icon-following-filled@2x.png").default,
+  home: require("./assets/icons/icon-home@2x.png").default,
+  homeFilled: require("./assets/icons/icon-home-filled@2x.png").default,
+}
```
maryseo070, line 76: oh in my profile-picture PR i have an icon table you can use once merged fyi

packages/fan-fix-ui-controls/TabBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+export const TabBar: React.FunctionComponent<TabBarProps> = (props) => {
+  const handleTabClick = (redirectPath?: string) => {
+    if (!redirectPath) {
+      return;
+    }
+    // TODO - EY
+    console.log('Redirecting to ', redirectPath);
```
maryseo070, line 103: any reason why you're not redirecting here? dont think it needs a todo but i may be missing something

maryseo070: ahh just saw your comment on using Link -- out of curiosity - why use the onClick in the first place/not just use link now?

packages/fan-fix-ui-controls/TabBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+    height: "100%",
+    maxWidth: 100,
+  }
+});
+
+const paths: Record<StatefulTabBarCategory, string> = {
+  home: "/home",
+  explore: "/explore",
+  following: "/following",
+  activity: "/activity",
```
maryseo070, line 63: nit: alpha sort

packages/fan-fix-ui-controls/TabBar.tsx
```diff
@@ -0,0 +1,138 @@
...
+    left: 0,
+    right: 0,
+    bottom: 0,
+    top: "auto",
+    position: "sticky",
+    paddingBottom: "20px",
+    backgroundColor: Palette.White,
+    zIndex: 1,
+  },
+  navBar: {
```
maryseo070, line 32: instead of duplicating this style code you can export that stylesheet from navbar and use those properties here

packages/fan-fix/HomeScreen.tsx
```diff
@@ -0,0 +1,144 @@
...
+@observer
+export class HomeScreen extends React.Component {
+  static inject: Inject<ReactElement> = () => {
+    return () => (
+      <HomeScreen />
+    );
+  }
+  
+  private readonly subscriptionList: []; // List<Subscription>; // TODO: replace once data model is available - EY
+  private readonly categoryList: []; // List<Category>; // TODO: replace once data model is available - EY
```
maryseo070, line 63: is the home screen supposed to be like the timeline feed? i dont think we will need these lists if that is the case - i think its best if we just remove tentative things like this and whoever implements them will add them in - that way they dont accumulate with no function!

maryseo070 (commented):
Dont forget to link the issue to your pr! even if its just the first part, its still nice to track! :)

packages/fan-fix-ui-controls/TabBar.tsx
adpoyao: I actually forgot to change the names here to tabBar and tabBarContainer. The css properties have changed from the original navBar :)

packages/fan-fix/HomeScreen.tsx
adpoyao: The home page is actually just screen 9 & 11. You could see that the little "home" icon on the bottom left is filled. Explore page is screen 10 and that's highlighted by the magnifying glass. I also aligned this with the way Kevin named his icon files for the tabNav.

I can remove the TODO here, but I have a computed value that relies on the subscriptionList, so that'll have to stay. I'll update it once my other PR with the data model is merged.

adpoyao:
Update: Seems like Borin's already built out the Bubble Chart (Category Circles)!
I'll replace the entire <BubbleChart> Component with his Component once he has his merged to master.

maryseo070 (approved):
looks good to me once you remove the D3 stuff and d3 from package.json!

packages/fan-fix-ui-controls/BubbleChart.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+import * as d3 from "d3";
+import $ from "jquery";
```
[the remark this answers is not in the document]
adpoyao: I'm actually going to remove BubbleChart entirely because I didn't realize Borin & Kevin have already created this category balls previously! :D

maryseo070:
@adpoyao i think you linked the wrong issue to this PR haha

PR #41, opened 2021-02-28
Profile picture functionality

maryseo070:
Sorry this one got ahead of me and became a bit long
- Creates image input component
- Creates photo models and functionality behind uploading photo
- Connects create profile page with uploading photo functionality
- Moves exco libs to shared-dependencies
- Creates Storage injectable, uses storage in fb storage model


packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -1,43 +1,46 @@
...
   ) {
     super();
-    this.appUserCollection = db.collection("AppUser")
-  }
-
-  public getAppUserRef(id: string): IdentifiableReference<AppUser> {
-    const userRef = new FirebaseObservableReference(
-      this.appUserCollection.doc(id),
+    this.dataManager = new FirebaseDataManager(
+      db.collection("AppUser"),
```
adpoyao, line 34: Just a refresher, my other PR has the "fetch" option implemented for FirebaseData Manager. Without "fetch", then it defaults to "realTime".

packages/fan-fix-data-model/firebase/FirebasePhotoModel.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(
+    db: UnifiedFirestoreTypes.Module,
+    private readonly appUserModel: FirebaseAppUserModel,
+    private readonly storage: firebase.storage.Storage,
+  ) {
+    super();
+  }
+
+  public async updateProfilePhoto(id: string, image: Image): Promise<void> {
+    this.appUserModel.updateAppUserProfilePhoto(await this.uploadPhoto("AppUser", id, image));
```
adpoyao, line 30: Curious, would we normally type out collectionName? In this case, for "AppUser" or is that overkill?

packages/fan-fix-ui-controls/package.json
```diff
@@ -7,7 +7,6 @@
     "test": "jest"
   },
   "dependencies": {
-    "mobx": "4.15.3",
     "mobx-react": "5.4.3",
```
adpoyao, line 10: Do we need to also remove mobx-react from fan-fix-ui-controls?

packages/fan-fix/AppInjector.ts
```diff
@@ -1,8 +1,7 @@
-import { Injector } from "@not-the-droids/exco-ts-inject";
+import { Injector, } from "@not-the-droids/exco-ts-inject";
```
adpoyao, line 1: Remove comma

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -1,43 +1,46 @@
...
     )
-    return userRef;
   }
+
+  public fetchAppUser(id: string): Promise<AppUser | undefined> {
+    return this.dataManager.fetch(id);
+  }
+
+  public async updateAppUserProfilePhoto(params: UpdateHasPhotoParams) {
+    await this.dataManager.upsert(params.id, serializeUpdateHasPhotoParams(params, this.staticClasses))
```
adpoyao, line 44: does this.staticClasses property exist anywhere in this class?

packages/fan-fix-data-model/firebase/FirebasePhotoModel.ts
maryseo070: umm im not sure if i understand the question. the upload photo function takes in the collectionName typed as string and then returns the promise. so the collection name "AppUser" is a proper string. am i missing what you're asking?

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
maryseo070: oops forgot to inject it! good catch

packages/fan-fix-data-model/firebase/FirebasePhotoModel.ts
adpoyao: Ah okay. What I meant was, do we usually have a place where we store the names of our collections as types, i.e.

`type Collection = "AppUser" | "Subscription"` etc and use that to type check instead of just string. I was merely curiously about it, not saying we need to implement this.

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,9 +1,10 @@
...
 import { AppUser } from "./AppUser";
+import { UpdateHasPhotoParams } from "./HasPhoto";
 
 export abstract class AppUserModel {
   static inject?: Inject<AppUserModel> = undefined;
 
-  public abstract getAppUserRef(id: string): IdentifiableReference<AppUser>
+  public abstract fetchAppUser(id: string): Promise<AppUser | undefined>;
+  public abstract updateAppUserProfilePhoto(params: UpdateHasPhotoParams): Promise<void>;
 }
```
adpoyao, line 10: I'm not sure what the convention is, but I wonder if it'd be more versatile to just have an `updateAppUserProfile` instead of limiting it to just only photo. The params would be pretty much everything in App user, but omitting properties such as isVerified, creatorStatus, createdSubscriptionList, followerStatus. What do you think?

maryseo070: hmm under different circumstances i would agree with you. but i think the case of updating the profile photo its a bit different. Since we have to serialize the parameters and since `FirebasePhotoModel`'s `updateProfilePhoto` function which handles uploading the photo to storage specifically takes in id an image properties, i think its enough of a special use case to warrant its own separate update function from the generic update user. also, if you check out how I'm calling updateAppUserProfilePhoto you'll see that it takes in the result of uploadPhoto - my opinion is that adding the additional conditions and checks that would be involved in just including this in an overarching update user function would be less clean

adpoyao: Okay, this makes sense! I'm about to create new methods for this to update AppUser profile with my FInish Profile functionality so I'll be able to trace it better. Thanks.

packages/shared-dependencies/package.json
```diff
@@ -11,6 +11,7 @@
     "firebase": "8.2.5",
     "react": "16.13.1",
     "react-dom": "16.13.1",
-    "mobx": "^4.15.3"
+    "mobx": "^4.15.3",
```
adpoyao, line 14: Remove ^

adpoyao (approved):

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -1,45 +1,52 @@
...
 
-  public getAppUserRef(id: string): IdentifiableReference<AppUser> {
-    const userRef = new FirebaseObservableReference(
-      this.appUserCollection.doc(id),
-      deserializeAppUser
-    )
-    return userRef;
+  public fetchAppUser(id: string): Promise<AppUser | undefined> {
+    return this.dataManager.fetch(id);
   }
```
mattryavec, line 47: Actually, because of Eddie's change to have `observeMode: "fetch"`, you can now safely return References and Lists from the DataManager! Use `initialize` to do the "fetch" or just wait for an observational context!

packages/fan-fix-data-model/firebase/FirebasePhotoModel.ts
mattryavec: Interesting approach Eddie.... In this case, we'd probably want an enum instead of a string union, since we're passing the value to the Firebase SDK, which would have no concept of our string union!

That said, my approach to these kinds of string constants is to avoid making constants by using them once (e.g., we pass in the collection name just once to the FirebaseDataManager).

HOWEVER, we do occasionally duplicate these names on the backend when we have to do something a bit different that wouldn't make sense to encapsulate behind a model. So I wouldn't be opposed to extracting out an enum of collection names to avoid duplication.

mattryavec (approved):
Spot on, Mary. Nailed it.

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
```diff
@@ -0,0 +1,45 @@
...
+export class FirebaseAssetModel extends AssetModel {
+  static inject: Inject<FirebaseAssetModel> = singletonInject((injector) => {
+    return () => new FirebaseAssetModel(
+      injector.get(FirestoreInjectable)(),
+      injector.get(FirebaseAppUserModel)(),
+      injector.get(FirebaseStorageInjectable)(),
+    )
+  });
+
+
```
adpoyao, line 20: nit: extra line

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
```diff
@@ -0,0 +1,45 @@
+import firebase from "firebase";
+import { FirestoreInjectable, UnifiedFirestoreTypes } from "@not-the-droids/exco-data-model-firebase";
+import { Inject, singletonInject } from "@not-the-droids/exco-ts-inject";
+import { UpdateHasPhotoParams } from "../HasPhoto";
+import { Image } from "../Image";
+import { AssetModel } from "../AssetModel";
+import { FirebaseAppUserModel } from "./FirebaseAppUserModel";
+import { FirebaseStorageInjectable } from "./FirebaseStorageInjectable";
+
```
adpoyao, line 9: nit: extra line

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,9 +1,9 @@
 import { Inject } from "@not-the-droids/exco-ts-inject";
-import { IdentifiableReference } from "@not-the-droids/exco-data-model";
 import { AppUser } from "./AppUser";
+import { UpdateHasPhotoParams } from "./HasPhoto";
 
 export abstract class AppUserModel {
   static inject?: Inject<AppUserModel> = undefined;
 
-  public abstract getAppUserRef(id: string): IdentifiableReference<AppUser>
+  public abstract updateAppUserProfilePhoto(params: UpdateHasPhotoParams): Promise<void>;
```
adpoyao, line 8: Will you need this if you have `updateProfilePhoto` in Asset data model?

maryseo070: yea, i think when you trace it more for your feature it will make sense - the updateProfilePhoto function calls this updateAppUserProfilePhoto passing in the results of uploadPhoto so we will need both.

PR #42, opened 2021-03-01
Post data model

maryseo070:
Basic post and app user action models


packages/fan-fix-data-model/AppUserAction.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface AppUserAction {
+  readonly appUser: IdentifiableReference<AppUser>;
+  readonly post?: IdentifiableReference<Post>;
+  readonly comment?: IdentifiableReference<Comment>;
+}
+
+export interface Comment extends AppUserAction {
+  readonly text: string;
+  readonly comments?: ReadonlyArray<Comment>;
+  readonly mentions?: ReadonlyArray<IdentifiableReference<AppUser>>;
```
mattryavec, line 14: Posts and Comments are going to have quite a few things in common. Perhaps extract out a `Postable` superinterface?

packages/fan-fix-data-model/AppUserAction.ts
```diff
@@ -0,0 +1,27 @@
+import { IdentifiableReference } from "@not-the-droids/exco-data-model";
+import { AppUser } from "./AppUser";
+import { Post } from "./Post";
+
+export interface AppUserAction {
+  readonly appUser: IdentifiableReference<AppUser>;
+  readonly post?: IdentifiableReference<Post>;
+  readonly comment?: IdentifiableReference<Comment>;
```
mattryavec, line 8: A `Postable` superinterface would also mean you could simply have the reference to the one object

packages/fan-fix-data-model/AppUserAction.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export interface Comment extends AppUserAction {
```
mattryavec, line 11: I think a Comment is less like an AppUserAction and more like a Post.

AppUserAction we may want to slim down to just the `appUser` and maybe some extra metadata (timestamp, e.g.). Then I think we'd want to include a `Post` as an AppUserAction, no?

mattryavec (approved):
I think we might want to chat briefly about this one! There are couple tweaks I think we should make to this model....

PR #69, opened 2021-03-02
11 finish profile functions

adpoyao:


packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,65 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+    const profileRef = this.appUserManager.collection.doc(userId);
+
+    await this.db.runTransaction(async (transaction) => {
+      const usernameSnapshot = await transaction.get(usernameRef);  
+
+      if (usernameSnapshot.exists && usernameSnapshot.data()?.uid !== userId) {
+        throw new Error(profileNameTaken);
+      }
+      
+      transaction.set(usernameRef, { uid: userId });
```
adpoyao, line 95: Question regarding setting username: should we go back and delete the old username from Username collection, or is that username not useable anymore?

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -1,18 +1,17 @@
...
   readonly realName?: RealName;
   readonly username?: string;
   readonly profilePhotoUrl?: string;
   readonly coverPhotoUrl?: string;
   readonly isVerified?: boolean;
   readonly bio?: string;
+  readonly websiteLink?: string;
   readonly creatorStatus?: CreatorStatus;
-  readonly createdSubscriptionList: List<Subscription>;
+  readonly createdSubscriptionList?: List<Subscription>;
```
maryseo070, line 14: hmm i could see why we would use lists for followed subscriptions since someone could follow a ton but for created subscriptions in my opinion an array of identifiable references would be good

my thoughts behind that is that the use case for grabbing these created subscriptions is when a creator is looking at their own profile and they see all their posts under their subscriptions. we would probably end up fetching posts that are contained in the array of subscription id's (through the identifiable ref) but wouldnt really need to get the value of the subscription object itself. 

will wait to see what matt says too hahah

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -32,8 +31,10 @@ export interface RealName {
   readonly last: string;
 };
 
+export type UpdateAppUserProfileParams = Omit<AppUser, "username" | "isVerified" | "creatorStatus" | "createdSubscriptionList" | "followerStatus">
```
maryseo070, line 34: can you move this to the AppUserModel? we usually keep our param type definitions in those model files

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,8 +1,19 @@
 import { Inject } from "@not-the-droids/exco-ts-inject";
 import { UpdateHasImageParams } from "./HasImage";
+import { UpdateAppUserProfileParams } from "./AppUser";
+import { UpdateUsernameParams } from "./Username";
 
+export const profileNameTaken = "Profile name taken";
```
maryseo070, line 6: huh...this sounds weird to me, shouldnt it be username taken?

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,65 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+  private readonly usernameManager: FirebaseDataManager<Username>;
   
   constructor(
-    db: UnifiedFirestoreTypes.Module,
+    private readonly db: UnifiedFirestoreTypes.Module,
     private readonly staticClasses: UnifiedFirestoreTypes.StaticClasses,
     ) {
       super();
-      this.dataManager = new FirebaseDataManager(
+      this.appUserManager = new FirebaseDataManager(
```
maryseo070, line 54: lol any particular reason why you changed this name? I kind of like the convention of calling it data manager!

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,65 @@ export class FirebaseAppUserModel extends AppUserModel {
...
         deserializeAppUser,
-        { observeMode: "fetch" }
+        { observeMode: "fetch" },
       );
+      this.usernameManager = new FirebaseDataManager(
+        db.collection("Username"),
+        deserializeUsername,
+        { observeMode: "fetch" },
+      )
+  }
```
maryseo070, line 64: ahh i just spoke to matt about something similar to this - would be a good opportunity to create a view model since username and appUser are kind of coupled together

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,65 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+    const username = sanitizeUsername(params.username);
+    
+    const usernameRef = this.usernameManager.collection.doc(sanitizeUsername(username));
+    const profileRef = this.appUserManager.collection.doc(userId);
+
+    await this.db.runTransaction(async (transaction) => {
+      const usernameSnapshot = await transaction.get(usernameRef);  
+
+      if (usernameSnapshot.exists && usernameSnapshot.data()?.uid !== userId) {
+        throw new Error(profileNameTaken);
```
maryseo070, line 92: hmm i feel a little weird throwing an error here. is it possible to just return undefined when unsuccessful, then if the username was successful? i just feel like throwing an error here requires someone calling the function to then wrap in a try/catch and handle the error and then they are limited with what they can do with it/may not even want to throw an error (maybe they just want to show a different message or something?). So my thinking is handling different return values is a bit more versatile. But maybe im off, thoughts?

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -1,13 +1,19 @@
 import React, { ReactElement } from "react";
 import { StyleSheet } from "aphrodite";
 import debounce from "lodash.debounce";
+import { History } from "history";
 import { observable, computed } from "mobx";
 import { observer } from "mobx-react";
+import { Authenticator } from "@not-the-droids/exco-auth";
 import { Inject } from "@not-the-droids/exco-ts-inject";
+import { HistoryInjectable } from "./HistoryInjectable";
+import { AppUserModel, profileNameTaken } from "../fan-fix-data-model/AppUserModel";
```
maryseo070, line 10: change to just fan-fix-data-model

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -85,31 +107,63 @@ export class FinishProfileScreen extends React.Component {
...
     const usernameIsValid = this.usernameMessage?.type === "success";
-    return !!(this.username && usernameIsValid && !!this.bio && !!this.websiteLink);
+    return !!(this.username && usernameIsValid && !!this.bio);
   }
 
   render() {
+
+    if (!this.props.authenticator.currentUser) {
+      // TODO: Route to Authentication Page
+    }
```
maryseo070, line 161: i dont think we'll need this here - later on we will add higher level routing stuff for if they are signed in

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
maryseo070: yeah my instinct is to delete it from the collection so that someone else can use it since its no longer actively being used. I would double check with kevins on this though just in case

packages/fan-fix-data-model/AppUser.ts
maryseo070: can you also remove the values you're not using in this PR? i.e.,) createdSubscriptionList - we cannot pass in lists as a param

packages/fan-fix-data-model/Username.ts
```diff
@@ -0,0 +1,11 @@
+import { Identifiable } from "@not-the-droids/exco-data-model";
+
+export interface Username extends Identifiable {
+  uid: string,
+};
+
+export interface HasUsername {
+  username: string; 
+}
```
maryseo070, line 9: hmm i dont think this naming works for this use-case. We're not really checking if they _have_ the username so i think just Identifiable & { username: string } is good!

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,65 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+  }
+
+  public async updateUsername(params: UpdateUsernameParams): Promise<void> {
+    const userId = params.id;
+    const username = sanitizeUsername(params.username);
+    
+    const usernameRef = this.usernameManager.collection.doc(sanitizeUsername(username));
+    const profileRef = this.appUserManager.collection.doc(userId);
+
+    await this.db.runTransaction(async (transaction) => {
```
maryseo070, line 88: does the firebase data manager `upsertWithTransaction` not work for this case?

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,65 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+      if (usernameSnapshot.exists && usernameSnapshot.data()?.uid !== userId) {
+        throw new Error(profileNameTaken);
+      }
+      
+      transaction.set(usernameRef, { uid: userId });
+      transaction.set(profileRef, { username }, { merge: true });
+    })
+  }
+  
+  public async updateAppUserProfile(params: UpdateAppUserProfileParams): Promise<void> {
```
maryseo070, line 100: i think `updateAppUser` would suffice as a name

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,8 +1,19 @@
...
 
+  public abstract isUsernameTaken(username: string): Promise<boolean>;
+  public abstract updateAppUserProfile(params: UpdateAppUserProfileParams): Promise<void>;
   public abstract updateAppUserProfilePhoto(params: UpdateHasImageParams): Promise<void>;
+  public abstract updateUsername(params: UpdateUsernameParams): Promise<void>;
+
+  public async updateFinishUserProfile(params: UpdateUsernameParams & UpdateAppUserProfileParams): Promise<void> {
+    await this.updateUsername(params);
+    await this.updateAppUserProfile(params);
+  }
```
maryseo070, line 18: this doesnt really belong..haha so the `AppUserModel` is defining all the abstract functions that should belong in it. So later on in, say, `FirebaseAppUserModel` where those functions are implemented, it knows what the params need to be and its return type as well as what functions need to exist. So this is more of an execution rather than an abstraction if that makes sense. Just like the one on line 13 - that one is correct

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -85,31 +107,63 @@ export class FinishProfileScreen extends React.Component {
     this.usernameMessage = message;
   }
 
+  onFinishClick = async () => {
+    this.error = undefined;
+    this.loading = true;
+
+    try {
+      await this.props.appUserModel.updateFinishUserProfile({
```
maryseo070, line 115: im not quite following how this works - i dont see `updateFinishUserProfile` defined in your firebase model so there must be some compile errors at least - maybe we should have a call? i think my comments are starting to be a bit much

maryseo070 (commented):
Plz link issue to PR. I think it might be helpful to do another call haha lmk if you want to

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
adpoyao: I like that you're thinking about versatility and not having to wrap the function calls in try/catch. In this case, I do want to throw a this specific custom error because then we are being explicit about this function not completing due to this verbose condition. I think if we return undefined/value, it's not very obvious to me what undefined means, and having a return value itself may not be meaningful inside the context of an update function in general. My opinion is that it's a good practice to have all functions in the top of the call stack be in a try/catch block to handle errors anyway, and this custom error will allow them to cherry pick it to be handled differently, or be treated as a general error if no specific handling is required.

packages/fan-fix/FinishProfileScreen.tsx
adpoyao: We were able to chat about this this morning - updateFinishUserProfile lives in AppUserModel and is being extended to the firebase model :)

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
maryseo070: haha i see your point..hmm we seem to be at a stalemate! loll @mattryavec thoughts?

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
adpoyao: For this one, I specifically only want to update what pertains to a user profile, so email, realName, bio, websiteLink (hence the omission of isVerified, creatorStatus, followerStatus, profilePhotoUrl, username... etc). This way, there's a method in the model that only allows user to update their textual profile stuff and not their status or anything that requires special handling (i.e. username and photo). What do you think?

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
mattryavec: My $0.02: Good idea

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
mattryavec: That one is more for transactionally updating a _single_ document managed by the data manager. Since we're using two different documents to enforce the uniqueness constraint, I recommended we go closer to the metal....

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,8 +1,21 @@
 import { Inject } from "@not-the-droids/exco-ts-inject";
+import { AppUser } from "./AppUser";
 import { UpdateHasImageParams } from "./HasImage";
+import { UpdateUsernameParams } from "./UsernameModel";
 
+export const usernameTaken = "Username taken";
+
+export type UpdateAppUserProfileParams = Omit<AppUser, "username" | "isVerified" | "creatorStatus" | "followerStatus" | "profilePhotoUrl" | "coverPhotoUrl">
```
mattryavec, line 8: Um.... it might be easier to just `Pick` the ones you want than `Omit` the ones you don't!

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,8 +1,21 @@
...
 export abstract class AppUserModel {
   static inject?: Inject<AppUserModel> = undefined;
 
+  public abstract isUsernameTaken(username: string): Promise<boolean>;
   public abstract updateAppUserProfileImage(params: UpdateHasImageParams): Promise<void>;
+  public abstract updateFinishProfile(params: UpdateAppUserProfileParams): Promise<void>;
```
mattryavec, line 14: "updateFinishProfile" just doesn't sound right. Also, it doesn't match your params interface, which isn't fun! I'd suggest just calling this `updateAppUserProfile`

packages/fan-fix-data-model/UsernameModel.ts
```diff
@@ -0,0 +1,5 @@
+import { Identifiable } from "@not-the-droids/exco-data-model";
+
+export type UpdateUsernameParams = Identifiable & {
+  username: string;
+};
```
mattryavec, line 5: Delete this file and move this to AppUserModel

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,8 +1,21 @@
...
+  public abstract updateUsername(params: UpdateUsernameParams): Promise<void>;
+
+  public async updateFinishUserProfile(params: UpdateUsernameParams & UpdateAppUserProfileParams): Promise<void> {
+    await this.updateUsername(params);
+    await this.updateFinishProfile(params);
+  }
```
mattryavec, line 20: I'm becoming less convinced this is the place for this. The only reason these two things are grouped this way, or that this this method even has "finish" in the name, is based on how the UI is laid out. Let's just have the UI call these underlying methods consecutively itself

packages/fan-fix-data-model/AppUserModel.ts
```diff
@@ -1,8 +1,21 @@
 import { Inject } from "@not-the-droids/exco-ts-inject";
+import { AppUser } from "./AppUser";
 import { UpdateHasImageParams } from "./HasImage";
+import { UpdateUsernameParams } from "./UsernameModel";
 
+export const usernameTaken = "Username taken";
```
mattryavec, line 6: I'd be a bit more descriptive with this variable name. My first thought is that this was a UI string and not a way to encode the type of error. I might suggest both making the name more descriptive and the string more... code-y?
```
const usernameTakenErrorMessage = "usernameTaken";
```

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +44,75 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+    await this.db.runTransaction(async (transaction) => {
+      const usernameSnapshot = await transaction.get(usernameRef);
+      const profileSnapshot = await transaction.get(profileRef);  
+
+      if (usernameSnapshot.exists && usernameSnapshot.data()?.uid !== userId) {
+        throw new Error(usernameTaken);
+      }
+      
+      const usernameId = usernameSnapshot.exists && usernameSnapshot.data()?.id;
+      const lastUsername = profileSnapshot.exists && profileSnapshot.data()?.username;
```
mattryavec, line 96: `currentUsername` ? Since it hasn't been changed yet? :)

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +44,75 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+
+    await this.db.runTransaction(async (transaction) => {
+      const usernameSnapshot = await transaction.get(usernameRef);
+      const profileSnapshot = await transaction.get(profileRef);  
+
+      if (usernameSnapshot.exists && usernameSnapshot.data()?.uid !== userId) {
+        throw new Error(usernameTaken);
+      }
+      
+      const usernameId = usernameSnapshot.exists && usernameSnapshot.data()?.id;
```
mattryavec, line 95: The id of the document doesn't live in the `data`. It would be on the snapshot itself (`usernameSnapshot.id`). And you already know the document exists because of line 91.

So really, you'd want to change line 98 to
```
if (sanitizeUsername(lastUsername) !== usernameSnapshot.id) {
```

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +44,75 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+      
+      if (lastUsername !== usernameId) {
+        const lastUsernameRef = this.usernameManager.collection.doc(lastUsername);
+        const lastUsernameSnapshot = await transaction.get(lastUsernameRef);
+        if (lastUsernameSnapshot.exists && lastUsernameSnapshot.data()?.uid === userId) {
+          transaction.delete(lastUsernameRef);
+        }
+      }
+      transaction.set(usernameRef, { uid: userId });
+      transaction.set(profileRef, { username }, { merge: true });
```
mattryavec, line 106: I don't thing we should be saving with the sanitized version of the username. Let them keep the casing they chose. The sanitized version is only for the uniqueness checking

packages/fan-fix/FinishProfileScreen.tsx
```diff
@@ -85,31 +107,58 @@ export class FinishProfileScreen extends React.Component {
     this.usernameMessage = message;
   }
 
+  onFinishClick = async () => {
+    this.error = undefined;
+    this.loading = true;
+
+    try {
```
mattryavec, line 114: @maryseo070 @adpoyao I would _strongly_ suggest using the [CallHandler](https://github.com/not-the-droids/coffee-now/blob/master/packages/coffee-now/ui/CallHandler.ts) I recently wrote for CN, especially since you all are in the early stages.

(Not for this PR, do it in a follow up)

packages/fan-fix-data-model/AppUserModel.ts
adpoyao: got it!

packages/fan-fix-data-model/AppUser.ts
mattryavec: I'm assuming this is for subscriptions that a creator creates. With our current specs, there will currently be only one of these. I'm good with the list (but don't make it optional).

Even if that changes, there would be few. Followers, on the other hand, could be following many many creators. For that, I can see still see a list, but you may want that to be a PaginatedList.

packages/fan-fix-data-model/Username.ts
```diff
@@ -0,0 +1,5 @@
+import { Identifiable } from "@not-the-droids/exco-data-model";
```
mattryavec, line 1: I'm being super picky, but I might move this file to the `firebase` folder. It's only used for Firebase-specific stuff, and isn't something we want or need exposed at the pure data model layer.

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -28,21 +43,85 @@ export class FirebaseAppUserModel extends AppUserModel {
...
+
+      if (usernameSnapshot.exists) {
+        if (usernameSnapshot.data()?.uid !== userId) {
+          throw new Error(usernameTakenErrorMessage);
+        }
+      
+        const currentUsername = profileSnapshot.exists && profileSnapshot.data()?.username;
+        
+        if (sanitizeUsername(currentUsername) !== usernameSnapshot.id) {
+          const currentUsernameRef = this.usernameManager.collection.doc(currentUsername);
```
mattryavec, line 107: You should be using the sanitized version as the id for the document reference.

packages/fan-fix-data-model/AppUser.ts
mattryavec: Taking on the actionable thing to the review: Don't make this optional.

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -7,19 +7,34 @@ import {
...
 
 const deserializeAppUser: DeserializeData<AppUser> = (id, data) => {
   return {
     id,
+    bio: data.bio,
     email: data.email,
+    username: data.username,
+    websiteLink: data.websiteLink,
     creatorStatus: data.creatorStatus,
     createdSubscriptionList: data.createdSubscriptionList,
```
mattryavec, line 23: This.... won't work. You need to construct an actual `List`. If you're not ready for this yet, delete the `createdSubscriptionList ` property for now.

adpoyao: I'm removing this property for now. Will include it once we're ready for it.

mattryavec (approved):

PR #72, opened 2021-03-03
Screen size styling, Icon update, dependency & ui components reshuffle

adpoyao:


packages/fan-fix-data-model/index.ts
```diff
@@ -20,4 +20,7 @@ export * from "./firebase/FirebaseSubscriptionModel";
 export {
   FirestoreStaticClassesInjectable,
   FirestoreInjectable,
-} from "@not-the-droids/exco-data-model-firebase";
\ No newline at end of file
+} from "@not-the-droids/exco-data-model-firebase";
+
+// others
+export * from "./HistoryInjectable";
```
adpoyao, line 26: I moved "history" dependency to be in shared-dependecies package; moved HistoryInjectable to data-model so it could be accessed by ff-admin too. Let me know if there's a better place to put this file.

packages/fan-fix-ui-controls/IconButton.tsx
```diff
@@ -23,27 +23,32 @@ const localStyles = StyleSheet.create({
...
   },
   gradient: {
-    marginBottom: 24,
     display: "flex",
     justifyContent: "center",
     alignItems: "center",
     cursor: "pointer",
+    borderRadius: "50%",
+    backgroundSize: "contain",
+    backgroundImage: `url(${require("./assets/icons/create-button-bg.png").default})`,
```
adpoyao, line 51: Instead of importing the circles 3 times, I asked Kevin Haag for the background image and making it the default background for gradient buttons once. This way, we have more versatility to controlling the styling of the shape and borders instead of being stuck to a circle image.

packages/fan-fix/AuthScreen.tsx
```diff
@@ -5,15 +5,15 @@ import { computed, observable } from "mobx";
...
+import {
   Palette,
-  Screen,
   StyledButton, 
   StyledText,
   StyledTextInput,
   View,
 } from "fan-fix-ui-controls";
-import { HistoryInjectable } from "./HistoryInjectable";
+import { Screen } from "./Screen";
```
adpoyao, line 16: Moved Screen from ff-ui-control to ff package. I think this isn't as granular as <View> and <StyledTextInput>. This also allows Screen and its children NavBar & TabNav to have direct access to the shared dependencies already laid out in webpack.config.js of fan-fix package. I think Coffee Now also has Screen placed in the coffee-now package (under ui folder)

adpoyao:
This one's a general update on past shared ui components - not tied to specific issue.

packages/fan-fix/TabBar.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const getStatefulIcon = (category: StatefulTabBarCategory): JSX.Element => {
+    const iconState = iconStates[category];
+    const icon = props.activeCategory === category ? iconState.filled : iconState.default;
+    const redirectPath = props.activeCategory === category ? undefined : paths[category];
+    return (
+      <IconButton
+        source={icon}
+        iconSize="small"
+        styles={localStyles.icon}
+        onClick={() => redirectPath ? console.log('Redirect to: ', redirectPath) : {}}
```
adpoyao, line 85: I have a question regarding history injection implementation. I'm going to keep the functionalities separate here and keep going with a new ticket and will implement this in next iteration.

packages/fan-fix-ui-controls/IconButton.tsx
maryseo070: oh sweet! Do we still need gradientBackground then? Seems like theres a bit of overlap

packages/fan-fix-ui-controls/style/DarkIcon.ts
```diff
@@ -1,20 +1,38 @@
 export type DarkIcon = 
+"activity" |
+"activityFilled" |
```
maryseo070, line 3: do these filled ones have a special background like that gradient? Just wondering if we could just use a backgroundColor attribute instead of adding these as assets

packages/fan-fix/CreateProfileScreen.tsx
```diff
@@ -4,11 +4,12 @@ import { StyleSheet } from "aphrodite";
 import { History } from "history";
 import { Authenticator } from "@not-the-droids/exco-auth";
 import { Inject } from "@not-the-droids/exco-ts-inject";
-import { IconButton, Palette, Screen, StyledButton, StyledText, View } from "fan-fix-ui-controls";
+import { AssetModel, FirebaseAssetModel, Image } from "fan-fix-data-model";
+import { observable } from "mobx";
+import { IconButton, Palette, StyledButton, StyledText, View } from "fan-fix-ui-controls";
```
maryseo070, line 9: nit: sort plzz

maryseo070: i know this was my screen haha i fuqed up sorting

packages/fan-fix/TabBar.tsx
```diff
@@ -0,0 +1,162 @@
+import React, { ReactElement } from "react";
+import { StyleSheet } from "aphrodite";
+import { History } from "history";
+import { IconButton } from "../fan-fix-ui-controls/IconButton";
+import { View } from "../fan-fix-ui-controls/View";
```
maryseo070, line 5: my vsc does this too its so annoying - can you move all these to just fan-fix-ui-controls?

packages/fan-fix/TabBar.tsx
```diff
@@ -0,0 +1,162 @@
...
+        iconSize="small"
+        styles={localStyles.icon}
+        onClick={() => redirectPath ? console.log('Redirect to: ', redirectPath) : {}}
+      />
+    )
+  }
+
+  render() {
+
+    console.log('111', this.props);
```
maryseo070, line 105: delete

packages/fan-fix-ui-controls/style/DarkIcon.ts
adpoyao: That's a good idea, I will check with Kevin. When I implemented these, the icons were not updated yet in Figma.

packages/fan-fix-ui-controls/IconButton.tsx
adpoyao: Oops! Forgot to delete. Thanks!

maryseo070 (approved):

packages/fan-fix-ui-controls/IconButton.tsx
maryseo070: oh i am using this component in ImageInput - can you double check to make sure it still works with your updates?

PR #73, opened 2021-03-03
Temporary creator application screen and functionality

maryseo070:
- removes all fake implementations
- adds temporary application for creator (just pushes button and sets acceptedAt)
- create app user view model


packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -20,6 +20,22 @@ const deserializeAppUser: DeserializeData<AppUser> = (id, data) => {
...
+    data.email = appUser.email;
+  }
+  if (appUser.profileImageUrl) {
+    data.profileImageUrl = appUser.profileImageUrl;
+  }
+  if (appUser.creatorStatus) {
+    data.creatorStatus = {
+      [appUser.creatorStatus]: staticClasses.FieldValue.serverTimestamp()
+    }
+  }
```
mattryavec, line 35: Careful here. Applying to become a creator is a bit more than a simple update to one's AppUser object! (I mean, it's basically that, but bear with me.)

First, you probably don't want to update the rest of the user's profile _and_ apply to be a creator at the same time. Conceptually they are disjoint operations (even if you _can_ accomplish both in one Firestore document update).

Second, we need to think about our security here. We don't want the user to be able to update their status to whatever they want! Otherwise a clever person could just set themselves to accepted! Another reason to keep this operation separate so we have better control over how this is implemented (e.g., maybe we decide later this is better handled as a call to a cloud function rather that a direct document update)

maryseo070: ah yeah i thought when we stubbed out the issues -  for this ticket we discussed it should just be a makeshift temporary button that allows us to become  a creator really quickly and create posts. did i msunderstand?

mattryavec: Just to be clear, I'm not suggesting we need to go to a cloud function at all right now! Just that you should make a separate method here on the AppUserModel. (`creatorApply` or something)

packages/fan-fix-data-model/AppUser.ts
```diff
@@ -27,6 +25,8 @@ export interface CreatorStatus extends AppUserStatus {
   rejectedAt?: Date;
 };
 
+export type CreatorStatusType = "appliedAt" | "acceptedAt" | "rejectedAt";
```
adpoyao, line 28: Would you want to capture AppUserStatus here too: suspended and deactivated?

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -42,7 +58,15 @@ export class FirebaseAppUserModel extends AppUserModel {
       );
   }
 
+  public getAppUserRef(id: string): IdentifiableReference<AppUser> | undefined {
+    return this.dataManager.getRef(id);
```
adpoyao, line 62: Just a headsup, `dataManager` variable has changed up appUserManager in my PR #69 to make it distinct between username and appUser collections.

packages/fan-fix/AppUserViewModel.ts
```diff
@@ -0,0 +1,23 @@
+import { computed } from "mobx";
+import { IdentifiableReference } from "@not-the-droids/exco-data-model";
+import { Inject, singletonInject } from "@not-the-droids/exco-ts-inject";
+import { Authenticator } from "@not-the-droids/exco-auth";
+import { AppUser, AppUserModel } from "fan-fix-data-model";
+
+export class AppUserViewModel {
```
adpoyao, line 7: Really liking the idea of a View Model 👍

packages/fan-fix/CreateProfileScreen.tsx
```diff
@@ -65,7 +67,11 @@ export class CreateProfileScreen extends React.Component<Props> {
...
   saveProfilePicture = async() => {
     try {
-      await this.props.assetModel.updateProfileImage(this.props.authenticator.currentUser?.id!, this.image);
+      const currentUserId = this.props.authenticator.currentUser?.id;
+      // create new app user based on id from authenticator
+      await this.props.appUserModel.updateAppUser({ id: currentUserId!, email: this.props.authenticator.currentUser?.providerData.emailAddress});
+      // set app user's profile photo
+      await this.props.assetModel.updateProfileImage(currentUserId!, this.image);
       this.props.history.push("/finishProfile");
     } catch(err) {
```
adpoyao, line 76: Note: I started keeping track of a loading and error state in the screens I'm building. I think it'll be worthwhile to start implementing those as you're making screens and functions so it's less of a chore to go back and add it to every method and screens later on.

packages/fan-fix/CreatorApplicationScreen.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return () => (
+      <CreatorApplicationScreen
+        appUserModel={injector.get(AppUserModel)()}
+        appUserViewModel={injector.get(AppUserViewModel)()}
+      />
+    );
+  }
+
+  sendApplication = async() => {
+    await this.props.appUserViewModel.currentAppUserRef?.initialize();
```
adpoyao, line 25: Just for my own learning, could we alternatively do:
`await when(() => this.props.appUserViewModel.currentAppUserRef.initialized)`?

packages/fan-fix-data-model/AppUser.ts
maryseo070: umm this will suffice for now. I'm only adding a temporary function to allow us to become creators while building out the rest of the project so we can handle suspending and deactivating separately another time.

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -42,7 +53,23 @@ export class FirebaseAppUserModel extends AppUserModel {
       );
   }
 
+  // TODO move to cloud function
```
mattryavec, line 56: Sorry for the confusion here. We may not need to do this. I just wanted the separate `creatorApply` method so that we left ourselves open to changing the implementation _if necessary_.

packages/fan-fix/CreatorApplicationScreen.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const currentUser = this.props.appUserViewModel.currentAppUserRef?.value;
+    await this.props.appUserModel.creatorApply(currentUser?.id!);
```
mattryavec, line 27: You _could_ create a `creatorApply` method on the ViewModel with zero arguments that would then forward onto the Model's creatorApply with the current user ID that you're getting from the ViewModel.

packages/fan-fix/package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@not-the-droids/exco-auth": "^0.4.2",
     "@not-the-droids/exco-auth-firebase-web": "0.0.1",
+    "@types/history": "4.7.8",
```
mattryavec, line 17: Pretty sure this should go in shared deps?

mattryavec (approved):
Just nits

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
maryseo070: ok haha noted

packages/fan-fix/package.json
maryseo070: ah yeah actually eddie added it to his PR that moves history to shared-deps so i will remove from here.

PR #75, opened 2021-03-04
Minimal create post funtionality

maryseo070:
- Create bare-bones create post function (no media implementations yet)


packages/fan-fix-data-model/PostModel.ts
```diff
@@ -1,5 +1,13 @@
 import { Inject } from "@not-the-droids/exco-ts-inject";
+import { Post } from "./Post";
+
+export type UpdatePostParams = Pick<Post, "isExclusive"> & {
+  id?: string,
+  subscriptionId: string,
```
aceiii, line 6: Had to dig a bit to get an understanding of how this `subscriptionId` works. Please correct me if I have something wrong.
 - `AppUser` (who are creators) creates a `Subscription`
 - `AppUser creates a `Post` with that `subscriptionId` to tie it to a `Subscription`
 - `Post` itself doesn't have a reference to the `AppUser` who created it except through `Subscription`
 - Will the creator have multiple `Subscriptions`?
 - Can an `AppUser` create a `Post` without a `Subscription`?
 - Is there a mechanism in place to prevent this `subscriptionId` from being arbitrarily selected? Say for example I enter the `subscriptionId` of another creator, it'll appear as if they made this Post

packages/fan-fix-data-model/PostModel.ts
```diff
@@ -1,5 +1,13 @@
...
+}
 
 export abstract class PostModel {
   static inject?: Inject<PostModel> = undefined;
+
+  public abstract updatePost(params: UpdatePostParams): Promise<void>;
```
aceiii, line 12: Is there no `createPost()`? I can see it's being used for both create and update below but I think it's more clear to have a separate method.

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -0,0 +1,46 @@
...
+      />
+    );
+  }
+
+  createPost = async() => {
+    const currentUserId = await this.props.appUserViewModel.currentAppUserRef?.id;
+    const subscription = await this.props.subscriptionModel.getSubscriptionByCreator(currentUserId!);
+    const subscriptionId = subscription[0]?.id;
+    await this.props.postModel.updatePost({
+      id: undefined,
```
aceiii, line 31: As mentioned above, I think a separate `createPost` would be clearer as it doesn't require `id` whereas `updatePost()` would and would catch that error if it's missing.

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
```diff
@@ -1,13 +1,61 @@
...
+const serializePost = (post: UpdatePostParams, staticClasses: UnifiedFirestoreTypes.StaticClasses) => {
+  const data: any = {};
+  if (post.isExclusive) {
+    data.isExclusive = post.isExclusive;
+  };
+  if (post.subscriptionId) {
+    data.subscriptionId = post.subscriptionId;
+  }
+  data.publishedAt = staticClasses.FieldValue.serverTimestamp();
+  return data;
```
aceiii, line 23: Weird, this ends up only serializing 3 properties?

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -3,9 +3,8 @@ import { AppUser } from "./AppUser";
 import { Category } from "./Category";
 
 export interface Subscription extends Identifiable {
-  categoryRefs: ReadonlyArray<IdentifiableReference<Category>>;
   creatorRef: IdentifiableReference<AppUser>;
-  description: string;
   price: number;
-  // posts: List<Post> // TODO: pending Post - EY
+  description?: string;
+  categoryRefs?: ReadonlyArray<IdentifiableReference<Category>>;
```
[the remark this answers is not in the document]
maryseo070: ahh yeah so i thought it should be optional because a subscription doesnt necessarily need to have categories associated with it. I can set it to empty though, that makes sense to me

packages/fan-fix-data-model/PostModel.ts
maryseo070: yeah so everything you wrote out is correct. The subscription is tied to the creator, not the post. Right now for V1 theyre expecting only 1 subscription per creator but we're just setting it to allow multiple for the future. i didnt add anything so far to prevent the user from setting another user's subscription id - i suppose moving this to a cloud function might be more secure. What do you think?

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
maryseo070: oh its only using the 3 properties i am using so far, as we build it out more we should include more properties but right now i just done have access to some of those values/types so i just skipped over them until theyre implemented

packages/fan-fix-data-model/PostModel.ts
mattryavec: Agreed.

packages/fan-fix-data-model/PostModel.ts
mattryavec: Mary, I think I got in your head about these cloud functions. There is a _LOT_ we can do to secure things just via Firestore's security rules, including allowing only the creator for a subscription to create or edit Posts with that subscription tied to it.

Cloud functions are useful when (a) it's too obtuse or simply impossible to get the security you need for an operation using just Firestore rules, (b) we have functionality that needs to be run server side (e.g., Stripe integrations), or (c) we simply want to structure an operation so it can be updated with _only_ a backend push rather than relying on a client-side update (which while easier with web can still at times result in old cached client-side code being used).

(Btw, that list is likely not exhaustive!)

packages/fan-fix-data-model/Subscription.ts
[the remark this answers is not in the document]
mattryavec: Yeah, I like the empty array rather than undefined as well!

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
```diff
@@ -1,13 +1,61 @@
...
 
 
+const serializePost = (post: UpdatePostParams, staticClasses: UnifiedFirestoreTypes.StaticClasses) => {
+  const data: any = {};
+  if (post.isExclusive) {
+    data.isExclusive = post.isExclusive;
+  };
+  if (post.subscriptionId) {
+    data.subscriptionId = post.subscriptionId;
+  }
```
mattryavec, line 21: These are not optional params. No need for the existence check.

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
```diff
@@ -1,13 +1,61 @@
...
+  data.publishedAt = staticClasses.FieldValue.serverTimestamp();
```
mattryavec, line 22: This will cause the publishedAt date to change whenever the post is updated, not just on create. You might want to leave the standard serialization params here and move the operation specific items to the method itself:

```
upsert(id, { ...serializePost(params), publishedAt: staticClasses.FieldValue.serverTimestamp() });
```

packages/fan-fix-data-model/SubscriptionModel.ts
```diff
@@ -6,4 +6,5 @@ export abstract class SubscriptionModel {
   static inject?: Inject<SubscriptionModel>;
 
   public abstract getSubscriptionRef(id: string): IdentifiableReference<Subscription>;
+  public abstract getSubscriptionByCreator(creatorId: string): Promise<ReadonlyArray<Subscription>>;
```
aceiii, line 9: Are we returning an array because a creator may create multiple subscriptions in the future?
If so, adding the 's' to hint at the plurality of the result from `getSubscriptionByCreator` would help.

aceiii (approved):
Looks good to me

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -31,18 +39,26 @@ export class FirebaseSubscription extends SubscriptionModel {
...
 
   public getSubscriptionRef(id: string): IdentifiableReference<Subscription> {
     return this.subscriptionDataManager.getRef(id);
   }
+
+  public getSubscriptionByCreator(creatorId: string): Promise<ReadonlyArray<Subscription>> {
+    return this.subscriptionDataManager.fetchArray((collection) => {
+      return collection
+      .where("creatorId", "==", creatorId)
+      .limit(1)
```
mattryavec, line 61: Nit: indent

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -31,18 +39,26 @@ export class FirebaseSubscription extends SubscriptionModel {
   }
 
   private deserializeSubscriptionData: DeserializeData<Subscription> = (id, data) => {
-    const { creatorRef, description, price } = data;
+    const { creatorId, description, price, categoryIds } = data;
+    const creatorRef = this.appUserModel.getAppUserRef(creatorId);
```
mattryavec, line 43: Nit: Inline like your categoryRefs?

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -0,0 +1,45 @@
...
+        postModel={injector.get(PostModel)()}
+        subscriptionModel={injector.get(SubscriptionModel)()}
+      />
+    );
+  }
+
+  createPost = async() => {
+    const currentUserId = await this.props.appUserViewModel.currentAppUserRef?.id;
+    const subscription = await this.props.subscriptionModel.getSubscriptionByCreator(currentUserId!);
+    const subscriptionId = subscription[0]?.id;
```
mattryavec, line 29: I think getting the creator's subscription will be a fairly common operation, and I'd hate to spread out this "get the first element" logic everywhere. It might behoove to move this to the `AppUserViewModel` (e.g., `fetchCreatorSubscription`).

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Inject } from "@not-the-droids/exco-ts-inject";
+import { observer } from "mobx-react";
+import { AppUserViewModel } from "./AppUserViewModel";
+import { Screen, StyledButton, View } from "fan-fix-ui-controls"
+import { PostModel, SubscriptionModel } from "fan-fix-data-model";
+
+interface Props {
+  appUserViewModel: AppUserViewModel;
+  postModel: PostModel;
+  subscriptionModel: SubscriptionModel;
```
mattryavec, line 11: DELETE!

mattryavec (approved):

PR #76, opened 2021-03-05
Post & Post Detail: Components Building

adpoyao:


adpoyao:
This is the first iteration before the PR gets too long. Even though it's the Post Detail page ticket that I'm working on, this PR mainly deals with creating post-related components that can be used through out the app:
- PostHeader (used for Home page, Post Detail, and Explore Pages)
- Post (used for Home page and Post Detail)
- Added Light Icon to Icon Library
- Added Empty Post Detail Screen & Expands on Home Screen

The actual functionality is NOT done. I had to hard feed in some props that could later be injected once the data model and functionality is all set, so I would focus solely on component architecture and styling for this PR. Thanks!

Next Iteration: Expand on Post component to be resusable in Post Detail Page + further enhancements to said Screen pages.

packages/fan-fix/Post.tsx
```diff
@@ -0,0 +1,153 @@
...
+          />
+        </View>
+
+        {/* Body: Post Content */}
+        <View styles={style.postContent}>
+          <View>
+            <StyledText variant="body2" styles={style.caption}>good morning love bugs! this post is for all...</StyledText>
+            {
+              captionLink && (
+                <StyledText variant="body2" styles={style.captionLink}>{captionLink}</StyledText>
```
maryseo070, line 129: what is this captionLink? I dont remember having a link in the post model

packages/fan-fix/PostDetailScreen.tsx
```diff
@@ -0,0 +1,53 @@
...
+const tabBarProps: TabBarProps = {
+  activeCategory: undefined,
+}
+
+@observer
+export class PostDetailScreen extends React.Component<Props> {
+  static inject: Inject<ReactElement> = (injector) => {
+    return () => (
+      <PostDetailScreen
+        history={injector.get(HistoryInjectable)()}
```
maryseo070, line 37: curious what the history will be needed for? seems like once youre on the post detail screen you can navigate with the navbars.

packages/fan-fix/PostHeader.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Icon, CircleImage, Palette, StyledText, View } from "fan-fix-ui-controls";
+
+type Size = "default" | "large";
+type ColorVariant = "dark" | "light";
+interface Props {
+  colorMode?: ColorVariant,
+  name: string,
+  profileImageUrl: string,
+  size?: Size,
+  username: string,
```
maryseo070, line 12: can you move the required on top and optional on bottom?

packages/fan-fix-ui-controls/StyledImage.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { css } from "aphrodite";
+import { mergeStyles, StylableProps } from "./Stylable";
+
+interface Props extends React.DetailedHTMLProps<React.ImgHTMLAttributes<HTMLImageElement>, HTMLImageElement>,
+  StylableProps {}
+
+export const StyledImage: React.FunctionComponent<Props> = (props) => {
+  const { styles, ...rest } = props;
+  return (
+    <img {...rest} className={css(mergeStyles(styles))} />
```
maryseo070, line 11: there's no inherent style that goes along with this? like border radius and height and stuff?

packages/fan-fix/Post.tsx
```diff
@@ -0,0 +1,153 @@
...
+          <StyledImage src={backgroundImage} styles={style.backgroundImage} />
+        )}
+
+        {/* Header: Post Info */}
+        <View styles={style.postInfo}>
+          <PostHeader
+            profileImageUrl={profileImageUrl}
+            name="charli d'amelio"
+            username="charlidamelio"
+            colorMode={colorMode}
```
maryseo070, line 119: this post header is this thing, right?
image
if so, i think we can rename it to something separate from post and have it handle its own data since i see it being used in the creator profile page, in the subscriptions list and under search results for explore as well - maybe something like AppUserSummaryTag or something? it seems like its not specific to the creator since in account settings the app user sees their own little header with their circle image and username and stuff as well. Idk im bad at naming im sure you can come up with something better than that.

maryseo070: Also i think it would be better for this component to get its data on its own instead of the parent passing it down

packages/fan-fix/Post.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+          {/* Footer: Actions */}
+          <View styles={style.postActions}>
+            <View styles={style.stats}>
+              <View styles={style.stat}>
+                  <Icon mode={colorMode} source="heartOutline" styles={style.statIcon}/>
+                  <StyledText styles={style.statText}>34.3K</StyledText>
+              </View>
+              <View styles={style.stat}>
+                <Icon mode={colorMode} source="comment" styles={style.statIcon}/>
```
maryseo070, line 143: These icons should be buttons, correct? With some type of action associated with their onClick? So FYI in my current PR i am making a StyledTouchableOpacity component that you could probbaly use here and wrap the icon in it so it has an onClick

packages/fan-fix/Post.tsx
adpoyao: You're right, I thought a link is something separate a user could share. I've removed this an changed it to a <span> instead within the <StyledText>

packages/fan-fix/PostDetailScreen.tsx
adpoyao: Yup, I'll remove it!

packages/fan-fix-ui-controls/StyledImage.tsx
adpoyao: There's not inherent style atm, but now that you're pointing this out, I'm going to just remove this file and use <image> in the Post component direction.

packages/fan-fix/HomeScreen.tsx
```diff
@@ -72,7 +73,12 @@ export class HomeScreen extends React.Component {
                 </View>
             </View>
           ) : (
-            <View>TBD</View>
+            <View>
+              {/* TODO: Post component should be able to determine what variant a post. For sake of screen development, I'm hard feeding it a prop temporarily. */}
+              <Post variant="caption"/>
```
adpoyao, line 78: TODO: Create a <Posts> component. Comment updated in upcoming PR.

packages/fan-fix/Post.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { Icon, Palette, StyledText, View } from "fan-fix-ui-controls";
+import { UserHeader } from "./UserHeader";
+import { Tag } from "./Tag";
+
+type Variant = "caption" | "media";
+interface Props {
+  variant: Variant;
+}
+
+const localStyle = {
```
mattryavec, line 12: I don't understand why this is called `localStyle`. Why have these been broken out this way?

packages/fan-fix/Tag.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { StyleSheet } from "aphrodite";
+import { Palette, StylableProps, StyledText, View } from "fan-fix-ui-controls";
+
+type ColorVariant = "dark" | "light";
+interface Props extends React.HTMLAttributes<HTMLDivElement>, StylableProps {
```
mattryavec, line 6: Why expose all of the div props?

helloeddiepo: Good point, I'll remove this

packages/fan-fix/Post.tsx
helloeddiepo: In this PR, I am using the  element and the style for that jsx needed to be separate from StyleSheet.create(...), but it was a bad case of naming on my part. In my PR #82, I will have a new StyledImage component is created in ff-ui-controls along with a BackgroundImage reusable component so the styling won't have to be broken out. I believe this issue will be resolved in that PR.

packages/fan-fix/Post.tsx
```diff
@@ -0,0 +1,158 @@
...
+    paddingLeft: 8,
+    fontSize: 10,
+    lineHeight: "10px",
+    fontWeight: 500,
+    letterSpacing: 1,
+    textTransform: "uppercase",
+  },
+});
+
+export const Post: React.FunctionComponent<Props> = (props) => {
```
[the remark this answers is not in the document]
helloeddiepo: FYI: I'm also renaming Posts to PostsView but in a subsequent PR

mattryavec (approved):

PR #78, opened 2021-03-08
Image posts

maryseo070:
- Create image posts
- Saves images to firebase storage under /Post/id
- Creates styledTouchableOpacity to use in Image Input so we can use file opener with different buttons 

NOTE: im not very confident in the whole uploadFile process/how i am handling the image type. Very open to suggestions on improvement.


packages/fan-fix-data-model/HasFile.ts
```diff
@@ -1,7 +1,11 @@
-export type PostFileType = "photo" | "pdf" | "text" | "video";
+import { Identifiable } from "@not-the-droids/exco-data-model";
+
+export type PostFileType = "image" | "pdf" | "text" | "video";
```
helloeddiepo, line 3: Is text considered a file type? From my understanding, a text is a string (and allows links) and is kept separate from an uploaded file.

aceiii: Sounds like "text" could be text files? Shouldn't the post already have the functionality to include text?

packages/fan-fix-ui-controls/ToggleButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+    transition:" right 0.2s ease-in-out",
+  },
+});
+
+export const ToggleButton: React.FunctionComponent<Props> = (props) => {
+  const { onClick } = props;
+  return (
+    <div onClick={onClick}>
+      <input type="checkbox" id="key" className={css(localStyles.input)}/>
+      <label htmlFor="key" className={css(localStyles.label)}>
```
helloeddiepo, line 47: I believe the key here for input id and label htmlFor should be dynamic.

packages/fan-fix-ui-controls/style/DarkIcon.ts
```diff
@@ -14,6 +15,7 @@ export const DarkIcons: { [T in DarkIcon]: string} = {
   avatar: require("../assets/icons/avatar.png").default,
   ellipsis: require("../assets/icons/darkEllipsis.svg").default,
   email: require("../assets/icons/email.svg").default,
+  lock: require("../assets/icons/lock.png").default,
```
helloeddiepo, line 18: Just heads up in my other PR, I've added another folder layer here: /assets/icons/dark/...

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -1,43 +1,135 @@
...
+  onSrcUpdate = (src?: string) => {
+    this.src = src;
   }
 
   render() {
     return (  
       <Screen>
-        <View>
+        <View styles={styles.root}>
+          <StyledTextInput placeholder="What's on your mind?" onChange={(event) => this.caption = event.target.value}/>
```
helloeddiepo, line 108: I'd make the onChange function into a method so you could expand on things such as data validations.

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -1,43 +1,135 @@
...
+            <ImageInput image={this.image} onChange={this.onImageChange} onSrcUpdate={this.onSrcUpdate}>
+              <View styles={styles.mediaContainer}>
+                <Icon styles={styles.icon} source="add" size="small"/>
+                <View styles={styles.mediaText}>Attach Media</View>
+              </View>
+            </ImageInput>
+          {
+            this.image && this.src && (
+              <View styles={styles.photoContainer}>
+                <img src={this.src} height={"auto"} width={"100%"} />
```
helloeddiepo, line 118: What's the convention for styling elements that don't have a Styled version in ff-ui-control? I prefer not doing the styling inline and use `style={}` and point to an object, probably next to the existing styles object. I think I had a similar question in my existing PR too; I gave the interim name of `localStyle` but I'm sure I'll have to change it up.

packages/fan-fix/CreatePostScreen.tsx
helloeddiepo: Note: I've made some changes to StyledTextInput to accomodate this style by giving it a variation of "comment" for my "Post Detail" screen, but I see it's being used in more than just comment section. I'll rename that variation to something more universal and push it in upcoming PR so you could use that here too.

packages/fan-fix-ui-controls/ToggleButton.tsx
maryseo070: oh right haha i got this from a snippet kevin sent me so forgot to update that, good catch

packages/fan-fix/CreatePostScreen.tsx
maryseo070: oh this PR is not focused on styling/elements its just functionality so i wll have to go back to style the screen. For stuff like that is there an issue using css function?

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
```diff
@@ -5,39 +5,48 @@ import { UpdateHasImageParams } from "../HasImage";
...
+    if (typeof file === "string") {
+      return { id, url: file, type };
     }
-    const ref = this.storage.ref(`${collectionName}/${id}/photo`);
-    await ref.put(image);
-    const imageUrl: string = await ref.getDownloadURL();
-    return { id, imageUrl };
+    const ref = this.storage.ref(`${collectionName}/${id}/image`); // todo update for videos and pdf
+    await ref.put(file);
+    const url: string = await ref.getDownloadURL();
```
aceiii, line 49: I don't think this `.getDownloadURL()` is what we want here even though it may look it. See https://firebase.google.com/docs/reference/js/firebase.storage.Reference#getdownloadurl

> Fetches a long lived download URL for this object.

This URL it returns is long-lived and would provide such functionality by including an access token in the url itself which I can see is the case in the files currently  on the dev project. I think we'll need to rethink this as we wouldn't want the url with the token directly stored as that would allow anyone able to pull in the post data to download all files.

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
```diff
@@ -5,39 +5,48 @@ import { UpdateHasImageParams } from "../HasImage";
...
   public async updateProfileImage(id: string, image: Image): Promise<void> {
-    this.appUserModel.updateAppUserProfileImage(await this.uploadImage("AppUser", id, image));
+    this.appUserModel.updateAppUserProfileImage(await this.uploadFile("AppUser", id, image, "image"));
   }
 
-  private async uploadImage(collectionName: string, id: string, image: Image): Promise<UpdateHasImageParams> {
-    if (!image) {
-      return { id };
+  public async updatePostFileImage(id: string, image: Image, type: PostFileType): Promise<void> {
+    this.postModel.updatePostFile(await this.uploadFile("Post", id, image, "image"));
```
aceiii, line 36: This doesn't seem like it's required to me and creates a dependency from to PostModel from wthin AssetModel. I think it woudl be clearer to have the AssetModel only handle the details related to the the file storage mechanism (uploading/delete/download/etc) and not let the caller deal with persisting the data to firestore separately

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
aceiii: Ok so more on this after some testing here are some observations:
 - `getDownloadURL()` returns an URL with an access token that allows the file to be accessed by anyone anywhere
 - the firebase storage security rules are checked when `getDownloadURL()` is called (security rules are bypassed with that access token)
 - firestore security rules also don't provide access control to specific fields in a document (the user reads the whole document or nothing, so storing these urls with their token on the Post document means the urls won't be secure)

I've come up with a couple ways of working around this:
1. Use the firebase store security rules to enforce who can access the file:
  - structure the firebase storage in a way where the subscription id is part of the file path
  - create cloud functions that set custom claims for users when their subscriptions change
  - write firebase storage security rules that checks if the subscription id in the filepath is in the users subscriptions claim
  - store only the firebase storage ref string (`gs://fan-fix-dev.appspot.com/Post/1xNJzXhyAx7S1nXlbTUs/image`) and use the sdk to create the ref and call `getDownloadURL()` at the time the image is viewed
  - there's a limit of 100 bytes for custom claims so would become limiting quite quickly
2. Use a subcollection on `Post` with a security rule to prevent reading files
  - create subcollection `Files` on `Post`
  - upload files to firebase storage, call getDownloadURL, store url with token in subcollection
  - write firestore security rule that checks if Post's subscriptionId is within users subscriptions field
  - firebase storage rules can be simpler to only allow the user to read/write to specific path

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
```diff
@@ -28,6 +32,25 @@ export class FirebaseAssetModel extends AssetModel {
...
+  }
+
+  // todo update parmas/refactor later to allow video as well
+  private async uploadFile(collectionName: string, id: string, file: Image, type: PostFileType): Promise<UpdateHasFileParams> {
+    if (!file) {
+      return { id, type };
+    }
+    if (typeof file === "string") {
```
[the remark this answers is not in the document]
maryseo070: the idea behind the string and the uploadFile function returning the id still is that the results of the uploadFile function get fed into the other update functions which take in the object's id and upsert the new data (i.e., image url or whatever) into the document - does that still seem off to you?

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
maryseo070: like move updatePostFileImage to the post model and just call uploadFile in there?

packages/fan-fix-data-model/firebase/FirebaseAssetModel.ts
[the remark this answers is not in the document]
aceiii: Yea it feels a bit off. I think the dependency between AssetModel and PostModel should be reversed:

`AssetModel` can be made more generic and doesn't need to care about how it gets used and doesn't need to know about `PostModel`. When `PostModel.updatePostFile()` is called it can check whether the file is a string and serialize it to the correct interface that Post expects files to be in. `PostModel` can be aware of `AssetModel` and know when it needs to call `AssetModel.uploadFile()` and can then serialize the result how it likes. The same change be made for `AppUserModel`/`updateProfileImage()` .

What happens when the direction you are going is that you end up repeating the same pattern of adding a function this model which is a simple wrapper on a call to the model with the result of uploadFile (You can see how `updatePostFileImage` and `updateProfileImage` are quite similar but don't have any real implementation.). This also adds a lot unecessary dependencies to this `AssetModel`.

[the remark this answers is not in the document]
maryseo070: oh that makes sense! thanks for explaining!

packages/fan-fix/CreatePostScreen.tsx
helloeddiepo: I created a <StyledImage> component in my upcoming PR, reason being that the image element is being used frequently in our app so it made sense to be able to pass in a style object from StyleSheet using that component. Make a note to use that component once my PR merges!

packages/fan-fix-data-model/AssetModel.ts
```diff
@@ -3,5 +3,5 @@ import { Image } from "./Image";
 export abstract class AssetModel {
   static inject?: Inject<AssetModel> = undefined;
 
-  public abstract updateProfileImage(id: string, image: Image): Promise<void>;
+  public abstract uploadFile(collectionName: string, id: string, file: Image): Promise<string>;
```
mattryavec, line 6: While I appreciate that this is more generic, this leaks implementation information out. Now UI code needs to know about Firebase collections! Bad!

It makes sense to have a single `uploadFile` private method in the FirebaseAssetModel that would accept a collection name that each public method would simply forward to.

I would suggest one of two things:
1. Stick with methods for each type of thing you want to upload: `uploadProfileImage`, `uploadPostVideo`, etc.
2. Go with your single method `uploadFile`, but use an enum to describe what type of file you're uploading rather than the raw collection name

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -145,9 +149,22 @@ export class FirebaseAppUserModel extends AppUserModel {
     return this.appUserManager.getRef(id);
   }
 
-  public async updateAppUserProfileImage(params: UpdateHasImageParams) {
-    await this.appUserManager.upsert(params.id, serializeUpdateHasImageParams(params, this.staticClasses))
-  } 
+  public async updateAppUserProfileImage(params: UpdateHasFileParams): Promise<void> {
+    const { url, id } = params;
+    let upsertData;
```
mattryavec, line 154: The common thing to slice of here is the `imageUrl`. How about updating your conditional to just extract the imageUrl?

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -54,6 +55,6 @@ export class FirebaseSubscription extends SubscriptionModel {
   }
 
   public getSubscriptionListByCreator(creatorId: string): List<Subscription> {
-    return new FilteredList(this.subscriptionDataManager.list, subscription => subscription.creatorRef.id === creatorId)
+    return new FilteredList(this.subscriptionDataManager.list, subscription => subscription.creatorId === creatorId)
```
mattryavec, line 58: Whoa whoa whoa. Too much copy-pasta! Let's think about what this will do. This fetches ALL subscriptions for ALL creators and _then_ filters that list client side! Use a QUERY!

The only time this is acceptable is when we're dealing with a limited amount of data and it makes sense and be more efficient to keep it all in memory.

packages/fan-fix-ui-controls/StyledTouchableOpacity.tsx
```diff
@@ -0,0 +1,47 @@
...
+    backgroundColor: "transparent",
+    width: "100%",
+    padding: 0,
+  },
+  touched: {
+    opacity: 0.5,
+  }
+});
+
+export const StyledTouchableOpacity: React.FunctionComponent<Props> = (props) => {
```
mattryavec, line 24: Is this supposed to be a generic touchable/pressable? Or is this a `StyledButton` since you're using a `<button>`?

packages/fan-fix-ui-controls/ToggleButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+});
+
+export const ToggleButton: React.FunctionComponent<Props> = (props) => {
+  const { id, onClick } = props;
+  return (
+    <div onClick={onClick}>
+      <input type="checkbox" id={id} className={css(localStyles.input)}/>
+      <label htmlFor={id} className={css(localStyles.label)}>
+        <div className={`toggleHandle ${css(localStyles.toggleHandle)}`}></div>
+      </label>
```
mattryavec, line 50: We don't want to deal with html ids in React. Wrap the entire thing in the `<label>` tag to achieve what you want.

https://reactjs.org/docs/forms.html

packages/fan-fix-ui-controls/StyledTouchableOpacity.tsx
maryseo070: its just a generic touchable that can wrap anything - should be a div instead? I used button because i added onMouseUp and onMouseDown events to show the opacity on click/touch and stuff

packages/fan-fix-ui-controls/ToggleButton.tsx
maryseo070: oh yeah i spoke to kevin about this component and he gave me a little code snippet to use so i was trying to use more css (i.e., the :checked + label styling -- should i revise this to just use different classnames based on react state?

maryseo070: ah i spoke to kevin about this component and we were trying to use more css but now that im reading that link you sent, makes sense to make react the single source of truth for state

mattryavec: Understood. But remember, CSS !== HTML. Stay in React-land as much as possible (and it's almost always possible!)

packages/fan-fix-ui-controls/StyledTouchableOpacity.tsx
mattryavec: Yeah, you probably want just a straight div. Those will also give you mouseDown and mouseUp events.

I also might rename this to just `StyledTouchable` or `StyledPressable`. The "opacity" part doesn't really apply!

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
```diff
@@ -51,9 +62,27 @@ export class FirebasePostModel extends PostModel {
     )
   }
 
-  public async createPost(params: CreatePostParams): Promise<void> {
-    await this.dataManager.upsert(undefined, { ...serializePost(params), publishedAt: this.staticClasses.FieldValue.serverTimestamp()});
+  public async createPost(params: CreatePostParams): Promise<string> {
+    const id = await this.dataManager.upsert(undefined, { ...serializePost(params), createdAt: this.staticClasses.FieldValue.serverTimestamp() });
+    const file = serializePostFiles(params);
+    if (file && file.type === "image") {
+      const url = await this.assetManager.uploadFile(`Post/${id}/File/`, id, file.url!);
```
mattryavec, line 69: Okay, let's remember Borin's synopsis here. We won't be able to store the URL, otherwise it will be publicly accessible. You'll have to somehow make a call to getDownloadUrl when a user accesses a post in order to download.

I would also suggest standardizing the names of these files. It will also help when Borin has to trigger asset processing. I would suggest something like `unprocessed_${fileId}` and then Borin can save the new thing as `${fileId}` and `blurred_${fileId}` e.g.

You need to coordinate with @aceiii on how you want to connect these dots.

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
```diff
@@ -6,23 +6,32 @@ import {
...
   data.subscriptionId = subscriptionId;
+  if (post.text) {
+    data.text = post.text;
+  }
   return data;
 }
 
+const serializePostFiles = (post: CreatePostParams) => {
+  return post.files ? post.files[0] : undefined; // files is an array for future cases but V1 will only handle single files per post
+}
```
mattryavec, line 28: Not quite sure the point of this, even for future proofing. You don't need to serialize these since you're just using it directly for your logic.

I'd suggest just iterating over the `files` in `createPost`

packages/fan-fix-data-model/Post.ts
```diff
@@ -1,8 +1,18 @@
 import { Identifiable, List } from "@not-the-droids/exco-data-model";
 import { Actionable } from "./Action";
-import { HasFile, PostFileType } from "./HasFile";
 import { Statistic } from "./Statistic";
+
+export type PostFileType = "image" | "pdf" | "video";
+
+export type HasFile<T extends PostFileType = PostFileType> = {
```
mattryavec, line 7: I don't believe this type is at all relevant anymore, no? You might end up with an `Asset` data object and you could put a `List<Asset>` on Post that would fetch from your `Post/id/File/` collection.

Either way you go, let's kill this for now.

packages/fan-fix-data-model/firebase/FirebaseAssetManager.ts
```diff
@@ -0,0 +1,25 @@
...
+    )
+  });
+
+  constructor(
+    db: UnifiedFirestoreTypes.Module,
+    private readonly storage: firebase.storage.Storage,
+  ) { }
+
+  public async uploadFile(path: string, id: string, file: File): Promise<string> {
+    const ref = this.storage.ref(`${path}`);
```
mattryavec, line 20: It's already a string. No need to template.

packages/fan-fix-data-model/PostModel.ts
```diff
@@ -1,16 +1,16 @@
 import { Inject } from "@not-the-droids/exco-ts-inject";
-import { Post } from "./Post";
+import { HasFile, Post, PostFileType } from "./Post";
 
-export type UpdatePostParams = Pick<Post, "isExclusive"> & {
-  id?: string,
+export type UpdatePostParams = Pick<Post, "isExclusive" | "text"> & {
+  id: string,
   subscriptionId: string,
+  files?: ReadonlyArray<HasFile<PostFileType>>,
```
mattryavec, line 7: Ah... okay, you're using HasFile here. Let's just define it here instead. Perhaps call it `UpdatePostFileParams` ?

packages/fan-fix-ui-controls/StyledTouchable.tsx
```diff
@@ -0,0 +1,47 @@
...
+    width: "100%",
+    padding: 0,
+  },
+  touched: {
+    opacity: 0.5,
+  }
+});
+
+export const StyledTouchable: React.FunctionComponent<Props> = (props) => {
+  const [touched, setTouched] = useState(false);
```
mattryavec, line 25: Is there any particular reason we're switching to Hooks over mobx? I suppose you might have wanted to keep the functional nature of a lower level component? It feels like a premature optimization to me, and at the expense of mixing how we manage component state in the app.

As far as performance goes, I've yet to see any real concrete evidence that functional components make a substantial impact. Especially when you have to recreate your closures for event handlers each time.

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
maryseo070: yeah, borin and i previously agreed to store the url under the File subcollection so we could have the security rules take care of hiding the url. I'll talk to borin more about how to name the files.

packages/fan-fix-data-model/PostModel.ts
maryseo070: oh i actually use it in Post.ts and i just realized i could just include files in my Pick for UpdatePostParams so i dont have to use it here anymore. but now that you mention the List of assets im rethinking this..

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
mattryavec: Aha! Good idea!

packages/fan-fix-ui-controls/StyledTouchable.tsx
maryseo070: oops yeah that was just a force of habit while trying to make these components really quick. switching back

packages/fan-fix-data-model/firebase/FirebasePostModel.ts
maryseo070: ok i spoke more with borin about it - he's adding all the preview data to an object called `preview` in the Post document which will be accessible to everyone since it will have the blurred version and information about processing progress. 

I'll be saving the files under the File subcollection which will be guarded by security rules. So once the post has been created and file uploaded, we'll just call the processing functions to pull from the subcollection so i dont believe we need to rework any naming.

packages/fan-fix-data-model/Post.ts
```diff
@@ -1,10 +1,20 @@
 import { Identifiable, List } from "@not-the-droids/exco-data-model";
 import { Actionable } from "./Action";
-import { HasFile, PostFileType } from "./HasFile";
 import { Statistic } from "./Statistic";
+
+export type PostFileType = "image" | "pdf" | "video";
+
+export type Asset<T extends PostFileType = PostFileType> = {
```
mattryavec, line 7: Still not quite sure what you have in mind for this generic parameter, but I could see it end up being handy. But it might be more trouble than its worth.

You could drop it for now, and always bring it back later. Since you an assign a default (as you've done here), anything you type as `Asset` would type as `Asset<PostFileType>` (e.g., you could change line 18 to just `List<Asset>` and it would be equivalent)

Something to consider :)

packages/fan-fix-ui-controls/ToggleButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+  constructor(props: Props) { 
+    super(props);
+  }
+  
+  @observable private checked: boolean = false;
+
+  private handleToggle = () => {
+    this.props.onClick();
+    this.checked = !this.checked;
+  }
```
mattryavec, line 57: Uh oh, you've duplicated state here. You maintain an internal state of the toggle and then do an external state of the toggle, both based on clicks. These states now have nothing to do with each other and might not be in sync. You also have no way to _set_ the checked value.

Let's make this a bound input, like our TextInputs. Add a prop for `checked` and an `onChange` that will callback with the new toggled value. That way you can do something like:

```
<ToggleButton
  checked={this.isExclusive}
  onChange={value => this.isExclusive = value}
/>
```

mattryavec (approved):
Just the thing with the toggle, then looks good.

PR #79, opened 2021-03-10
PDF & Video posts

maryseo070:
- Update post functions to accept different types (pdf, image, video) - only PDF and image are hooked up/tested. Video will be the next PR
- Rename image input to media input


helloeddiepo:
The _Media.ts_ file is empty. Do you need that for anything?

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -83,16 +83,21 @@ export class CreatePostScreen extends React.Component<Props> {
     const subscription = await this.props.appUserViewModel.fetchCreatorSubscription();
     const subscriptionId = subscription?.id;
     if (subscriptionId) {
+      let type: PostFileType;
```
helloeddiepo, line 86: You could declare type directly inside the conditional (L89) since it's only being used in that scope.

packages/fan-fix-data-model/HasFile.ts
```diff
@@ -4,9 +4,15 @@ import { Image } from "./Image";
 export type PostFileType = "image" | "pdf" | "video";
 
 export type HasFile<T extends PostFileType = PostFileType> = {
-  type: T,
+  type?: T,
```
mattryavec, line 7: Why optional? (This might get resolved if we eliminte the `HasFile` type entirely as mentioned on #78 )

mattryavec (commented):
This one looks good, but until #78 has been resolved I'm holding off on approval for now.

Also, you have an empty `Media.ts` file. (Not sure when Github took away file-level comments, but it's annoying!)

packages/fan-fix/CreatePostScreen.tsx
```diff
@@ -81,22 +81,27 @@ export class CreatePostScreen extends React.Component<Props> {
     const subscription = await this.props.appUserViewModel.fetchCreatorSubscription();
     const subscriptionId = subscription?.id;
     if (subscriptionId) {
+      const files: Asset<PostFileType>[] = [];
+      if (this.file && this.file instanceof File) {
+        let type: PostFileType;
+        type = MediaTypes[this.file.type];
```
mattryavec, line 87: Just `const type = MediaTypes[this.file.type]` ? You also might want to check for undefined

packages/fan-fix-data-model/PostModel.ts
```diff
@@ -8,7 +8,11 @@ export type UpdatePostParams = Pick<Post, "isExclusive" | "text"> & {
 };
 
 export type CreatePostParams = Omit<UpdatePostParams, "id">;
-
+export const MediaTypes: Record<string, PostFileType> = {
```
mattryavec, line 11: Nit: I like keeping constants and variables uncapitalized in order to reserve capitalization for types. So how about just `mediaTypes` here?

mattryavec (approved):

PR #81, opened 2021-03-10
Exclusive Posts Locked Views

helloeddiepo:


packages/fan-fix-ui-controls/CircleImage.tsx
```diff
@@ -21,8 +21,8 @@ const localStyles = StyleSheet.create({
     width: 24,
   },
   large: {
-    height: 112,
-    width: 112,
+    height: 48,
+    width: 48,
```
maryseo070, line 25: oh make sure this doesnt effect other pages - i think the create profile page needs 112.

packages/fan-fix-ui-controls/IconTextButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+  },
+  darkText: {
+    color: Palette.Black,
+  },
+});
+
+export const IconTextButton = (props: Props) => {
+  const { colorMode, disabled, iconSource, onClick, styles, text } = props;
+
+
```
maryseo070, line 62: nit - remove extra lines

packages/fan-fix-ui-controls/style/LightIcon.ts
```diff
@@ -1,8 +1,14 @@
 export type LightIcon = 
+"addToQueue" |
 "comment" |
-"heartOutline";
+"lock" |
+"heartOutline" |
+"subscribed";
```
maryseo070, line 6: hmm what is this icon? Can we name it something separate from the subscription functionality?

packages/fan-fix/PostView.tsx
```diff
@@ -146,6 +180,32 @@ export const PostView: React.FunctionComponent<Props> = (props) => {
 
       <View styles={[style.container, containerStyle]}>
         <View styles={[style.post, postStyle]}>
+          {
```
maryseo070, line 183: oh i just noticed on line 174 the format for condition is wrong -can you update that one?

packages/fan-fix/PostView.tsx
```diff
@@ -146,6 +180,32 @@ export const PostView: React.FunctionComponent<Props> = (props) => {
...
+            (isLocked && !captionOnly) && (
+              <View styles={style.subscribeCTA}>
+                <Icon source="lock" styles={style.lockIcon}/>
+                <StyledText styles={style.subscribeCTAText}>Subscribe to unlock</StyledText>
```
maryseo070, line 187: hmm so im not 100% sure how this will work but @aceiii is working on the image processing to blur it out after the post is created so based on if its exclusive or not we will have security rules/different queries. i believe once we fetch it, will just need to add the transparent lock thing and not the opacity but i would check with borin.

packages/fan-fix/PostView.tsx
```diff
@@ -146,6 +180,32 @@ export const PostView: React.FunctionComponent<Props> = (props) => {
...
+            (isLocked && !captionOnly) && (
+              <View styles={style.lockedPostOverlay}/>
+            )
+          }
+          
+          {
+            (!captionOnly && backgroundImage) && (
+              <BackgroundImage source={backgroundImage} />
+            )
+          }
```
maryseo070, line 208: just curious/thinking out loud - instead of having all the conditions one after another, would it be possible to pass in the post media to some type of media display component that would use more TS (i.e., depending on the PostFileType) be able to determine what to display? As an example, i remember in CN we had MenuItemCustomizationView

packages/fan-fix/PostView.tsx
```diff
@@ -172,26 +232,30 @@ export const PostView: React.FunctionComponent<Props> = (props) => {
...
+                    </View>
+                    <View styles={style.stat}>
+                      <Icon mode={colorMode} source="comment" styles={style.statIcon}/>
+                      {
+                        showStats && (
+                          <StyledText styles={style.statText}>2.4K</StyledText>
+                        )
+                      }
+                    </View>
+                  </View>
```
maryseo070, line 255: will this footer section be used in more than one place? wondering if it would be useful to make this its own component too or if thats too much. thoughts? maybe in that component we could just inject the app user view model and add a check on there if theyre a creator or not so we can display stats/isLocked based on that

packages/fan-fix/ScreenTabs.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const handleClick = (event: React.ChangeEvent<HTMLDivElement>, option: string): void => {
+    if (onClick) {
+      onClick(option);
+    }
+  }
+
+  return (
+    <View styles={[style.container, styles]}>
+      {options.map(option => {
```
maryseo070, line 50: nit: extra line after {

packages/fan-fix-ui-controls/style/LightIcon.ts
helloeddiepo: The "subscribed" icon is the person icon with the little check mark. The other version is "subscribe", which is with a + mark. I can add the word Icon to the end of the string, but I personally don't think it'll get convoluted with the subscription functionality since this is only the name of an icon, as opposed to the other scenario where I had "Post" as a file name/ component and it made more sense to rename it to PostView. Let me know.
Screen Shot 2021-03-15 at 9 00 31 AM
Screen Shot 2021-03-15 at 9 00 37 AM

packages/fan-fix-ui-controls/CircleImage.tsx
helloeddiepo: Okay, I'll revert the large attribute and add a new "medium" size here for the 48 x 48  👍

packages/fan-fix/PostView.tsx
helloeddiepo: I really like this idea! My original intent of having one PostView component is because the two variants "caption" and "media" share many of the same css attributes, with the differences being a toggle between having a background image, inverting the light-dark color scheme, and some additional css treatment. With styling alone, it made more sense to me to put them in one component, but I can also see the benefits of having it separated out as well. Before I jump to making the changes, @mattryavec what do you think?

packages/fan-fix/PostView.tsx
helloeddiepo: I like the suggestion 👍  Currently, this footer is not being used anywhere else outside of the PostView component, but since I already have a "CreatorHeador", it also makes great sense to have a footer component too. Plus, I think if we do separate out a "PostMediaView" out of the existing "PostView" component, then this will become reusable as well. Making the changes now.

packages/fan-fix/CreatorProfileScreen.tsx
```diff
@@ -1,25 +1,40 @@
...
+  captionLink: {
+    color: Palette.Highlight,
+  },
 });
 
 const tabBarProps: TabBarProps = {
   activeCategory: undefined,
 }
 
+const screenTabsProps = ["exclusive", "public"];
```
mattryavec, line 36: Nit: Add a type declaration to ensure we don't add non-LOA strings. (Also, I like making these things Readonly to ensure nobody mucks with these constants!)
```
const screenTabsProps: ReadonlyArray<LevelOfAccess> = ...
```

packages/fan-fix/CreatorProfileScreen.tsx
```diff
@@ -28,6 +43,14 @@ export class CreatorProfileScreen extends React.Component {
     );
   }
 
+  handleScreenTabClick = (option: LevelOfAccess) => {
+    if (!option) {
```
mattryavec, line 47: Why would option be undefined? PUT YOUR FAITH IN TYPESCRIPT

packages/fan-fix/EngagementBar.tsx
```diff
@@ -0,0 +1,63 @@
+import React from "react";
+import { StyleSheet } from "aphrodite";
+import { Icon, StylableProps, StyledText, View } from "fan-fix-ui-controls";
+
+type ColorVariant = "dark" | "light";
```
mattryavec, line 5: Let's try and be consistent with our naming. Before you named the type `-Color` but the prop `colorMode`. Here the type is `ColorVariant` and the prop is named `colorMode`.

Let's pick one and stick with it! (But I don't like `ColorVariant` as it conflates with our `Variant` types IMO)

packages/fan-fix/EngagementBar.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <View styles={style.stat}>
+          <Icon mode={colorMode} source="heartOutline" styles={style.statIcon}/>
+          {
+            showStats && (
+              <StyledText styles={style.statText}>34.3K</StyledText>
+            )
+          }
+        </View>
+        <View styles={style.stat}>
+          <Icon mode={colorMode} source="comment" styles={style.statIcon}/>
```
mattryavec, line 53: Oh ho! You're forwarding this on to the `Icon`! In this case, let's export the Icon's color mode type and use it here. We don't want to redefine it! Maybe even change the name of the prop to `iconColorMode` so folks know exactly what it's used for

packages/fan-fix/PostView.tsx
mattryavec: Yeah, in general I prefer small components. Can't happen all the time (e.g., Screens are usually just gonna be kinda big), but for stuff like this I think a break up is a pretty good idea.

packages/fan-fix/ScreenTabs.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <View styles={[style.container, styles]}>
+      {
+        options.map(option => {
+          const isActive = option === activeOption;
+          const tabStyle = isActive ? style.activeTab : null;
+          const textStyle = isActive ? style.activeTabText : null;
+          return (
+            <View key={option} styles={[style.tab, tabStyle]} onClick={(e) => handleClick(e, option)}>
```
mattryavec, line 56: You don't need to forward the event. Or even capture the event as an arg here! Just `handleClick(option)`!

packages/fan-fix/ScreenTabs.tsx
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+import { StyleSheet } from "aphrodite";
+import { Palette, StylableProps, StyledText, View } from "fan-fix-ui-controls";
+
+interface Props extends StylableProps {
+  activeOption: string,
+  onClick: (option: string) => void,
```
mattryavec, line 7: Nit: I might call this `onOptionSelected` :)

packages/fan-fix/CreatorProfileScreen.tsx
helloeddiepo: 😂😂😂

packages/fan-fix/EngagementBar.tsx
helloeddiepo: Got it, I'll go and fix all the areas where I have `ColorVariant` into `-Color`

packages/fan-fix/EngagementBar.tsx
helloeddiepo: Got it!!

packages/fan-fix/ScreenTabs.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <View styles={[style.container, styles]}>
+      {
+        options.map(option => {
+          const isActive = option === activeOption;
+          const tabStyle = isActive ? style.activeTab : null;
+          const textStyle = isActive ? style.activeTabText : null;
+          return (
+            <View key={option} styles={[style.tab, tabStyle]} onClick={handleClick(option)}>
```
mattryavec, line 57: Wait, this doesn't work does it? You should need `onClick={() => handleClick(option)}`. Right now you're setting onClick to undefined!

mattryavec (approved):

packages/fan-fix/ScreenTabs.tsx
helloeddiepo: Oops!

PR #82, opened 2021-03-10
Explore & Subscriptions Screen

helloeddiepo:


maryseo070 (approved):

packages/fan-fix-ui-controls/style/DarkIcon.ts
```diff
@@ -29,6 +32,8 @@ export const DarkIcons: { [T in DarkIcon]: string } = {
   add: require("../assets/icons/dark/add.png").default,
   arrowLeft: require("../assets/icons/dark/darkArrowLeft.svg").default,
   avatar: require("../assets/icons/dark/avatar.png").default,
+  clearInput: require("../assets/icons/dark/clear-input.svg").default,
```
mattryavec, line 35: Huh... why the mismatch of SVGs and PNGs? Feels like we should pick one or t'other... Let's ask @thekevbot !

packages/fan-fix/SearchInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+    height: 16,
+    width: 16,
+  },
+});
+
+
+export const SearchInput: React.FunctionComponent<Props> = (props) => {
+  const { onClear, onChange, styles, value } = props;
+
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
mattryavec, line 44: I might alter the `StyledTextInput` to have its `onChange` event return the underlying string rather than forwarding on the full event. Should save us a bit of effort! And in this case, you can just forward the onChange prop in SearchInput directly to the StyledTextInput!

(But let's do that in a follow up PR so you can do everything all at once and not muddle this one!)

mattryavec (approved):
Looks great! Just a couple non-blocking comments

packages/fan-fix/SearchInput.tsx
helloeddiepo: You got it! Refactor Friday it is.

PR #104, opened 2021-03-17
Basic Timeline Data Model

maryseo070:
- create basic timeline interface
- create get timeline list from app user's timeline subcollection


packages/fan-fix-data-model/Timeline.ts
```diff
@@ -0,0 +1,4 @@
+export interface Timeline {
+  postId: string;
```
mattryavec, line 2: Why not `postRef: IdentifiableReference<Post>`?

Does it have to do with creating a circular dependency between AppUserModel and PostModel? If so, there are solutions:
1. Make a TimelineModel that injects both
2. Inject providers instead of instances so you don't instantiate them at construction time, but access them later by invoking the provider. (This one isn't my favorite)

packages/fan-fix-data-model/firebase/FirebaseAppUserModel.ts
```diff
@@ -11,7 +19,8 @@ import { AppUser } from "../AppUser";
 import { AppUserModel, usernameTakenErrorMessage, UpdateAppUserProfileParams, UpdateUsernameParams, UpdateAppUserParams } from "../AppUserModel";
 import { FirebaseAssetModel } from "./FirebaseAssetModel";
 import { serializeUpdateHasImageParams } from "./Serialization";
-import { UpdateHasFileParams } from "HasFile";
+import { UpdateHasFileParams } from "../HasFile";
```
helloeddiepo, line 22: nit: sort

packages/fan-fix-data-model/firebase/FirebaseTimelineModel.ts
```diff
@@ -0,0 +1,52 @@
...
+  static inject?: Inject<FirebaseTimelineModel> = singletonInject((injector) => {
+    return () => new FirebaseTimelineModel(
+      injector.get(FirestoreInjectable)(),
+      injector.get(FirebasePostModel)(),
+    )
+  });
+
+  constructor(
+    private readonly db: UnifiedFirestoreTypes.Module,
+    private readonly postModel: PostModel,
```
mattryavec, line 29: FYI, I've slightly changed my tune here. Since we're already in Firebase land at this point, seems reasonable to just inject the Firebase version of any dependencies. I mean, sure, it saves you a single import, but sometimes it's also convenient if you need to expose a Firebase-specific method on one Firebase model for use on another Firebase model.

(BUT I still like naming the variable just `postModel` :) )

mattryavec (approved):

PR #109, opened 2021-03-18
Explore Page: Subscribe to publication functionality

helloeddiepo:


helloeddiepo:
Note:
Renamed components CreatorCard/s to PublicationCard/s
Added date-fns to shared-dependency
Resolved a few tsc error; will resolve rest in coming PR.

packages/fan-fix-data-model/Subscription.ts
```diff
@@ -4,7 +4,7 @@ import { Publication } from "./Publication";
 
 export interface Subscription extends Identifiable {
   appUserRef: IdentifiableReference<AppUser>;
-  publicationRef: IdentifiableReference<Publication>;
+  publicationRef: IdentifiableReference<Publication>,
```
maryseo070, line 7: nit: semicolon, not comma

packages/fan-fix-data-model/SubscriptionModel.ts
```diff
@@ -4,7 +4,7 @@ import { Subscription } from "./Subscription";
 
 export type UpdateSubscriptionStatusParams = Pick<Subscription, "id" | "active" | "endDate">;
 
-export type CreateSubscriptionParams = Omit<Subscription, "id">;
+export type CreateSubscriptionParams = Pick<Subscription, "appUserRef" | "publicationRef" | "price">;
```
maryseo070, line 7: shouldnt this still omit the id if we're creating?

packages/fan-fix-data-model/firebase/FirebaseSubscriptionModel.ts
```diff
@@ -13,18 +13,14 @@ import { Subscription } from "../Subscription";
...
-  const { active, appUserRef, endDate, price, publicationRef, startDate } = subscription;
+  const { appUserRef, price, publicationRef } = subscription;
   const data: any = {
-    active,
+    active: true,
     appUserId: appUserRef.id,
     price,
     publicationId: publicationRef.id,
-    startDate,
+    startDate: new Date(),
```
maryseo070, line 22: use the server timestamp (see app user model or post model)

packages/fan-fix/AppUserViewModel.ts
```diff
@@ -11,21 +11,31 @@ export class AppUserViewModel {
...
 
-  public async fetchCreatorSubscription(): Promise<Subscription | undefined> {
-    const subscriptionList = await this.subscriptionModel.getSubscriptionListByAppUser(this.currentAppUserRef?.id!)
-    return subscriptionList.values[0];
+  public async fetchCreatorPublication(): Promise<Publication | undefined> {
+    const publicationList = await this.publicationModel.getPublicationListByCreator(this.currentAppUserRef?.id!);
+    return publicationList.values[0];
+  }
+
+  public fetchAppUserSubscriptions(): List<Subscription> {
```
maryseo070, line 37: if you are updating/changing something, please check where it is used. this function is being used in the create post screen so changing it will need to be updated there too.

packages/fan-fix/BillingScreen.tsx
```diff
@@ -82,10 +82,10 @@ export class BillingScreen extends React.Component {
     this.loading = true;
 
     try {
-      console.log('Save button clicked');
+      console.log("Save button clicked");
       this.loading = false;
     } catch (error) {
-      console.error('Error updating billing information: ', error);
+      console.error("Error updating billing information: ", error);
```
maryseo070, line 88: instead of putting loading false in both try and catch you can just move it to the end outside of the block

packages/fan-fix/CreatorHeader.tsx
```diff
@@ -1,10 +1,14 @@
...
 import { StyleSheet } from "aphrodite";
+import { AppUser } from "fan-fix-data-model";
 import { Icon, CircleImage, Palette, StylableProps, StyledText, View } from "fan-fix-ui-controls";
+import { computed } from "mobx";
+import { observer } from "mobx-react";
 
 type Size = "default" | "large";
 type textColor = "dark" | "light";
 interface Props extends StylableProps {
+  creator?: AppUser,
```
maryseo070, line 11: shouldnt the creator be required if this whole header is based on it's data?

packages/fan-fix/CreatorHeader.tsx
```diff
@@ -37,30 +41,51 @@ const style = StyleSheet.create({
...
 });
 
+@observer
+export class CreatorHeader extends React.Component<Props> {
+  
+  @computed private get profileImageUrl(): string {
+    const { creator } = this.props;
+    const profileImageUrl = creator?.profileImageUrl|| ""; // TODO: Insert default avatar image if no url is available
+    return profileImageUrl;
+  }
```
maryseo070, line 51: my opinion is it would be nice to  make a creator model that gets this information and injects it so this component can get its own data and we can have the functions to get creator data to use elsewhere

packages/fan-fix/CreatorHeader.tsx
```diff
@@ -37,30 +41,51 @@ const style = StyleSheet.create({
...
 
-export const CreatorHeader: React.FunctionComponent<Props> = (props) => {
-  const { textColor, size, styles } = props;
+  @computed private get realName(): string | undefined {
+    const { creator } = this.props;
+    const firstName = creator?.realName?.first || "";
+    const lastName = creator?.realName?.last || "";
+    return `${firstName} ${lastName}`.trim() || undefined;
```
maryseo070, line 57: i would imagine this formatting would be used in multiple places - would be nice to extract so other places can use it as well (see CN RealName.ts)

packages/fan-fix/CreatorHeader.tsx
```diff
@@ -37,30 +41,51 @@ const style = StyleSheet.create({
...
+  }
 
-  // Dummy Data:
-  const name = "charli d'amelio";
-  const profileImageUrl = "https://pickaface.net/gallery/avatar/unr_sample_161118_2054_ynlrg.png";
-  const username = "charlidamelio"
+  @computed private get username(): string | undefined {
+    const { creator } = this.props;
+    const username = creator?.username || "";
+    return username ? `@${username}` : undefined;
```
maryseo070, line 63: same goes for username formatting

packages/fan-fix/CreatorHeader.tsx
```diff
@@ -37,30 +41,51 @@ const style = StyleSheet.create({
...
-      <CircleImage size={circleImageSize} withBorder={withBorder} imageSource={profileImageUrl} />
-      <View styles={style.userInfo}>
-        <View styles={style.row}>
-          <StyledText variant={headerStyle} styles={[style.realName, textStyle]}>{name}</StyledText>
-          <Icon source="verifiedCheck" styles={style.badgeIcon}/>
+    if (!creator) {
+      return null;
+    }
+
+    return(
```
maryseo070, line 78: nit: space after return

packages/fan-fix/CreatorHeader.tsx
```diff
@@ -37,30 +41,51 @@ const style = StyleSheet.create({
...
-  return (
-    <View styles={[style.container, textStyle, styles]}>
-      <CircleImage size={circleImageSize} withBorder={withBorder} imageSource={profileImageUrl} />
-      <View styles={style.userInfo}>
-        <View styles={style.row}>
-          <StyledText variant={headerStyle} styles={[style.realName, textStyle]}>{name}</StyledText>
-          <Icon source="verifiedCheck" styles={style.badgeIcon}/>
+    if (!creator) {
+      return null;
+    }
```
maryseo070, line 76: if creator is required for this component i dont think this condition would be necessary - lmk if there is something im missing about that but seems like it should be required.

packages/fan-fix/ExploreScreen.tsx
```diff
@@ -39,18 +29,64 @@ const tabBarProps: TabBarProps = {
...
 
-  @observable searchInput: string = "";
+  handleSubscribe = async (publicationRef: IdentifiableReference<Publication>): Promise<void> => {
+    const { appUserViewModel, subscriptionModel } = this.props;
+    try {
+      // TODO: Determine if user is creating a new subscription (done) or resubscribing to an expiring publication (todo) - EY
+      const createSubscriptionParams: CreateSubscriptionParams = {
+        appUserRef: appUserViewModel.currentAppUserRef!,
+        publicationRef,
+        price: publicationRef.value?.price!,
```
maryseo070, line 70: should we be exposing price like this? Im not an expert on security stuff (enter @aceiii) but since these are FE parameters arent they more vulnerable to being manipulated?

packages/fan-fix/ExploreScreen.tsx
```diff
@@ -39,18 +29,64 @@ const tabBarProps: TabBarProps = {
...
+      };
+      // TODO: Handle payment processing here before creating subscription
+      await subscriptionModel.createSubscription(createSubscriptionParams);
+      this.subscriptionList = await appUserViewModel.fetchAppUserSubscriptions();
```
maryseo070, line 74: dang we fetch the whole subscription list again after subscribing to one? i know we're not using observable lists another way to do this or it this ok @mattryavec ?

packages/fan-fix/ExploreScreen.tsx
```diff
@@ -39,18 +29,64 @@ const tabBarProps: TabBarProps = {
...
+
+    } catch (error) {
+      console.error("Error subscribing to selected publication", error);
+    };
+  }
+
+  handleUnsubscribe = async (publicationRef: IdentifiableReference<Publication>): Promise<void> => {
```
maryseo070, line 81: i know you're including this in the next iteration but just as a note - i dont believe the parameter will be a pulbication ref based on how its set up now. If you're getting a list of publications then the list will be actual pubs, not their references.

packages/fan-fix/PublicationCards.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { ScrollView } from "fan-fix-ui-controls";
+import { PublicationCard } from "./PublicationCard";
+import { Loading } from "./Loading";
+
+interface Props {
+  onSubscribe: (publicationRef: IdentifiableReference<Publication>) => void;
+  onUnsubscribe: (publicationRef: IdentifiableReference<Publication>) => void;
+  publicationList: List<Publication>;
+  subscriptionList?: List<Subscription>;
+}
```
maryseo070, line 16: why are we passing these down as props instead of injecting the models? i dont really like how the parent of the parent of the parent contains the subscribe/unsubscribe functions - thats a little hard to track and if we want to use those functions again would be nice to have them in some type of format where we can inject and use. Im also not convinced that we need this component all together - seems like it just takes a list of values and maps through them to show the pubilcation cards, sort of like the middle man right now passing props

